PR #232, opened 2018-11-10
fix: expose correct and functional Select and Card components as needed for skills dev

gcpantazis:
## Description

Please review individual commits, since the lint fixes pollutes the overall diff. @jtini I made those prettier changes to pass precommit verification.

@kengoldfarb this includes the fixes we were discussing, plus I figured out the various problems we were having with `Select`.

Thanks gang!

## Type

- [x] Feature
- [x] Bug
- [ ] Tech debt


packages/react-heartwood-components/src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -36,8 +36,8 @@ const ButtonGroup = (props: Props) => {
 								kind === 'floating'
 									? 'simple'
 									: kind === 'segmented'
-										? 'secondary'
-										: action.kind
+									? 'secondary'
```
gcpantazis: Unrelated to this prettier fixup but we probably should avoid nested ternaries, they're not very readable; there's a rule for that if we want to use it: https://eslint.org/docs/rules/no-nested-ternary

packages/spruce-next-helpers/src/skillskit/next/DevControls.js
```diff
@@ -42,16 +44,17 @@ class DevControls extends Component {
...
 					className={`select`}
 					label="Jump to Role"
 					onChange={this.onChangeRole}
-				>
-					<option value="">{`Current: ${auth.role}`}</option>
-					<option value="owner">Owner</option>
-					<option value="teammate">Teammate</option>
-					<option value="guest">Guest</option>
-				</select>
+					options={{
```
gcpantazis: @jtini @brandonminch it's probably worth mentioning that this API (using an object to define options) probably needs some expansion to specify default, etc., in case you ever want to use it in an uncontrolled manner

packages/spruce-next-helpers/package.json
```diff
@@ -72,6 +72,7 @@
 	"dependencies": {
 		"@babel/polyfill": "^7.0.0",
 		"@babel/register": "^7.0.0",
+		"@sprucelabs/react-heartwood-components": "7.8.1-pre-1be2f60.0",
```
gcpantazis: @kengoldfarb I had to specify a prerelease to get this to work for my skill; unsure if lerna handles upgrades to this JSON automagically (I think you said it did?)

packages/spruce-next-helpers/src/skillskit/next/DevControls.js
liquidg3: yo @gcpantazis - I was thinking you would set "value" or "defaultValue"  to handle selections.

packages/react-heartwood-components/src/components/ButtonGroup/ButtonGroup.js
liquidg3: I'm down for that, never said to myself, "I'm sure happy we have nested ternaries". thoughts @kengoldfarb ?

packages/spruce-next-helpers/src/skillskit/next/DevControls.js
gcpantazis: Discussed this with Tay on Slack but bringing it here just to close the loop:

Yeah! Wasn't really thinking about `defaultValue`. That's the main thing for uncontrolled, but options are super wonky, and sometimes you end up doing crazy stuff like:

`<option value="" selected disabled hidden>Choose here</option>`

... to get it to do exactly what you want. We don't need to handle that stuff right now, can cross that bridge when we come to it  😄

PR #236, opened 2018-11-16
Create canonical icon directory, automatically provide to react heartwood `Icon`

gcpantazis:
## Description

This does a few things:

1. Creates a canonical directory of design-approved icons in `heartwood-components`.
2. Automatically exposes their locations as a dictionary at buildtime (icons.json)
3. `react-heartwood-components` consumes this JSON, and uses it to build and transpile the icons into react-wrapped components (using babel-inline-svg)
4. Makes each SVG a string-specified option for `Icon` in `react-heartwood-components`, removing the prefix. i.e.:
    * `heartwood-components`: `ic_edit.svg`
    * `react-heartwood-components`: `<Icon icon="edit" isLineIcon />`
5. All available canonical options are displayed in storybook, under the knobs.

Notes:
* @jtini a few of the icon names were reserved JS names (delete, private, etc.). I renamed them in a commit below (three in total), but you may have better names!
* This commit does not migrate the existing custom icons to the canonicals, that should be done after we merge this. Custom `Icon`s still work in the manner we merged a few days ago.

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt


packages/heartwood-components/Gulpfile.js
```diff
@@ -1,31 +1,68 @@
-var gulp = require('gulp');
-var rename = require('gulp-rename');
-var sass = require('gulp-sass');
-var sassGlob = require('gulp-sass-glob');
-var concat = require('gulp-concat');
-var rename = require('gulp-rename');
-var uglify = require('gulp-uglify');
+var fs = require('fs-extra')
```
kengoldfarb: Not that it matters much in this case, but while we're in here, let's switch to `const` for consistency

gcpantazis: Done; enforcing that with `.eslintrc`.

PR #239, opened 2018-11-19
Feature/message component

jtini:
## Description

Adds the styling and markup for the Message component to be used in the Feed and Messages.

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt


packages/heartwood-components/components/message/message.hbs
```diff
@@ -0,0 +1,31 @@
+<div class="message">
+    <span class="message__pre">
+        {{#if from.image}}
+        {{#if from.isSprucebot}}
+        <img class="message__from-image{{#if isSprucebot}} message__from-sprucebot-image{{/if}}" src="{{ from.image.src }}"
```
brandonminch: @jtini looks like the `message__from-sprucebot-image` class isn't being applied

packages/heartwood-components/components/04-images-and-icons/01-avatar/avatar.config.js
```diff
@@ -15,35 +20,40 @@ module.exports = {
 			context: {
 				image
 			}
-		}, {
```
gcpantazis: Just wanna be sure `prettier`/`eslint` is configured properly here, I thought our ruleset kept opening braces on the same line if chained (could be wrong about that though)

packages/heartwood-components/public/stylesheets/global.css
```diff
@@ -4015,3 +4015,82 @@ select:-webkit-autofill:focus {
   @media screen and (min-width: 1000px) {
     .l-detail__content {
       padding: 0 2rem; } }
+
```
gcpantazis: Not to fix here, but we need to get build artifacts out of git history

packages/heartwood-components/components/message/message.hbs
jtini: fixed

packages/heartwood-components/components/04-images-and-icons/01-avatar/avatar.config.js
brandonminch: @gcpantazis looks like the prettier settings are the same as other projects. The vscode settings were not part of the project though, I have added the settings for format on save to the other PR I opened.

PR #248, opened 2018-12-08
Allow View and HeaderPrimary to set necessary search props

gcpantazis:
## Description

When implementing `View` this started throwing warnings, since the required props weren't being passed through.

I'm additionally removing the search-bar if those props aren't provided; I'd imagine we might want to use this view in a logged-in context without search.

## Type

- [ ] Feature
- [X] Bug
- [ ] Tech debt

### DIFF
--- a/packages/react-heartwood-components/src/components/Core/components/HeaderPrimary/HeaderPrimary.js
+++ b/packages/react-heartwood-components/src/components/Core/components/HeaderPrimary/HeaderPrimary.js
@@ -23,7 +23,13 @@ type Props = {
 	toggleSidebarVisibility: Function,
 
 	/** Set true to show the sidebar (small screens only) */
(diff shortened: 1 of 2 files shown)


packages/react-heartwood-components/src/components/Core/components/HeaderPrimary/HeaderPrimary.js
```diff
@@ -23,7 +23,13 @@ type Props = {
 	toggleSidebarVisibility: Function,
 
 	/** Set true to show the sidebar (small screens only) */
-	sidebarIsVisible: boolean
+	sidebarIsVisible: boolean,
```
brandonminch: @gcpantazis consider renaming to `isSidebarVisible` to follow convention we discussed.

PR #249, opened 2018-12-12
Brandon/skill stylesheet

brandonminch:
[SBL-1057](https://sprucelabsai.atlassian.net/browse/SBL-1057)

## Description

Proposal for skills linking to versioned stylesheet:

* Running `yarn build` in `heartwood-components` will output `heartwood-components.min.css` to the `/build` directory which the CI platform can deploy to a versioned directory behind a CDN i.e. `/stylesheets/HEARTWOOD_VERSION/heartwood-components.min.css`
* Updated `spruce-next-helpers/_document.js` to only output stylesheet set in the environment config. (Default will now come from the skill default config)
* Updated `spruce-skill/config/default.js` `SKILL_STYLESHEET` var to default to the CDN stylesheet with the current skill's version of `heartwood-components` so that it points to `https://CDN_URL/stylesheets/HEARTWOOD_VERSION/heartwood-components.min.css` @kengoldfarb see comment in the diff below.

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt


packages/spruce-skill/config/default.js
```diff
@@ -29,7 +30,10 @@ module.exports = {
 	METRICS_SEQUELIZE_DISABLED: process.env.METRICS_SEQUELIZE_DISABLED === 'true',
 	API_HOST: process.env.API_HOST,
 	API_KEY: process.env.API_KEY,
-	SKILL_STYLESHEET: process.env.SKILL_STYLESHEET,
+	SKILL_STYLESHEET:
+		process.env.SKILL_STYLESHEET ||
+		`https://CHANGE_ME_CDN_URL/stylesheets/${HEARTWOOD_VERSION ||
+			'latest'}/heartwood-components.min.css`,
```
gcpantazis: I'm assuming we'll put a cache on the CDN  so `latest` will run into issues there; I wonder if we should also hash the project for something like:

```
`https://${CDN_HOST}/stylesheets/heartwood-components-${HEARTWOOD_VERSION}-${HEARTWOOD_HASH}.min.css`
```

PR #251, opened 2018-12-14
fix(circleci) Add canary deploy separate from prerelease

gcpantazis:
- Keeps the `canary` tag preserved for the canary branch, which is what 
we merge into until we're ready to merge to dev (which is the npm beta 
tag.
- `prerelease/x` branches still publish for development purposes, but 
are not tagged.
- canary storybook/heartwood sites only updated on canary
- Upgrade docker to Node 10


.circleci/config.yml
```diff
@@ -2,22 +2,38 @@ version: 2
 jobs:
   test:
     docker:
-      - image: circleci/node:8
+      - image: circleci/node:10
     steps:
       - add_ssh_keys:
           fingerprints:
-            - 'ec:9f:2e:aa:1f:c9:ab:49:57:8f:c0:cd:2e:5b:f3:b0'
+            - "ec:9f:2e:aa:1f:c9:ab:49:57:8f:c0:cd:2e:5b:f3:b0"
```
kengoldfarb: Is your prettier not picking this up?  For me it changes it to single quotes when I save

gcpantazis: This is probably from the prettier settings being in vscode config. I'll fix.

PR #254, opened 2018-12-15
fix(heartwood): Remove min-height from .main-content

gcpantazis:
- It was set to 100vh, but that won't work correctly on mobile devices without JS intervention. So let's leave that off, and defer to the pages using that class to handle their heights if they desire.
- Sounds like y'all are deleting the CSS files so I'll need to re-push this once that's done with; just review the `scss` for now :)


gcpantazis:
Wrong branch 🤦‍♂️

PR #258, opened 2018-12-17
Add Table Components

jtini:
[DES-164](https://sprucelabsai.atlassian.net/browse/DES-164)

## Description

This adds table components to React Heartwood. These new components depend on [React Table](https://react-table.js.org/#/story/readme) and add opinionated structure and styling. I set this up so that a table may or may not have selectable rows, and can be optionally searchable, sortable, and filterable. This also includes <TableSearch /> and <TableFilters /> to add the correct styles.

A note on responsiveness; because we don't yet know all of the use cases for tables in skills and core, the responsive behavior is simple horizontal scrolling (this applies to situations where many filters are applied as well).

I'd also note that I haven't included the 'customize display' feature shown in the designs, since there isn't yet a table designed to use that feature.

Here's the [Figma link](https://www.figma.com/file/I0By1hIW5Y6sEkMUKkGa9dRz/Heartwood?node-id=2002%3A27517) for reference.

You may also notice that I added [Faker](https://github.com/marak/Faker.js/) and [uuid](https://www.npmjs.com/package/uuid) as dev dependencies to make it easier for us to generate dummy data when building these components.

## Type

- [•] Feature
- [ ] Bug
- [ ] Tech debt


packages/react-heartwood-components/package.json
```diff
@@ -63,6 +63,7 @@
 		"@storybook/addon-options": "^4.0.8",
 		"@storybook/addons": "^4.0.0-alpha.21",
 		"@storybook/react": "^4.0.0-alpha.21",
+		"Faker": "^0.7.2",
```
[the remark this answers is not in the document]
jtini: Updated

packages/react-heartwood-components/src/components/Table/Table.js
```diff
@@ -0,0 +1,276 @@
...
+			isSelectable,
+			kind,
+			pluralKind,
+			bulkActions,
+			sortable,
+			...rest
+		} = this.props
+		const { selectedIds, allRowsSelected } = this.state
+
+		let renderColumns = [...columns]
```
gcpantazis: `renderColumns` suggests a method to me; `columnsToRender`?

packages/react-heartwood-components/src/components/Table/Table-story.js
```diff
@@ -0,0 +1,184 @@
...
+		minWidth: 300,
+		maxWidth: 400,
+		Cell: row => <a href="#">{row.value}</a>
+	},
+	{
+		Header: 'Store',
+		accessor: 'storeNumber',
+		minWidth: 80,
+		maxWidth: 100
+		// width: '100%'
```
gcpantazis: Are these commented out for context or can they be removed?

packages/react-heartwood-components/src/components/Forms/components/Checkbox/Checkbox.js
```diff
@@ -36,13 +39,20 @@ export default class Checkbox extends Component<Props, State> {
 	}
 
 	handleChange = () => {
+		console.log('handleChange')
```
gcpantazis: Remove console.logs here and below

PR #259, opened 2018-12-19
SBL-1507: Fully Controlled Header

CarolynSM:
## Description

• Rename `sidebarIsVisible` to `isSidebarVisible` and `sidebarIsExpanded` to `isSidebarExpanded`
• Allow `<View>` to accept `toggleSidebarVisibility()`, `toggleSidebarExpanded()`, `isSidebarVisible`, and `isSidebarExpanded` props

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt


packages/react-heartwood-components/src/components/BigCalendar/BigCalendar.js
```diff
@@ -77,7 +77,7 @@ class BigCalendar extends Component<Props, State> {
 		doubleClickTime: 250,
 		doubleClickToCreate: false, // defaults to single click
 		eventTimeFormat: 'h:mma',
-		headerCellDowFormat: 'dd',
+		headerCellDowFormat: 'ddd',
```
CarolynSM: This is a residual fix for SBL-1495 that probably isn't worth it's own PR if we can help it - sneaking it in here

packages/react-heartwood-components/src/components/View/View.js
```diff
@@ -11,40 +11,51 @@ type Props = {
...
 	}
 
 	toggleSidebarVisibility = () => {
 		this.setState(prevState => ({
-			sidebarIsVisible: !prevState.sidebarIsVisible,
-			sidebarIsExpanded: true
+			isSidebarVisible: !prevState.isSidebarVisible,
+			isSidebarExpanded: true
 		}))
+		this.props.toggleSidebarVisibility && this.props.toggleSidebarVisibility()
```
gcpantazis: This isn't exactly a controlled component, since it maintains its own state here. More typical would be to call something like `onSidebarVisibilityChange`, and not keep internal state (only refer to the props, and defer to the downstream components to manage state). 

A challenge might be to both keep local state if `onSidebarVisibilityChange` was not provided, and otherwise use props if it was. So in that way you could have both a controlled and an uncontrolled mode. Necessary? I don't know. We really only *need* a controlled mode for the website.

gcpantazis:
Adding @brandonminch for review + consideration of my previous review comment

packages/react-heartwood-components/src/components/View/View.js
liquidg3: Confirming there is no point in having a stateful component here. Going fully controlled is the way to go!

packages/react-heartwood-components/src/components/View/View.js
```diff
@@ -11,79 +11,60 @@ type Props = {
...
+const View = (props: Props) => {
+	const {
+		sidebarItems,
+		user,
+		business,
+		getSearchSuggestionValue,
+		renderSearchSuggestion,
+		isSidebarVisible,
+		isSidebarExpanded,
+		setSidebarExpanded,
```
gcpantazis: Ah so. I think in react-heartwood this prop should actually be `onToggleSidebarExpanded`, since from this component's perspective it's a simple callback to alert downstream components that the user has requested the bool to flip. In other words, at this level it's not "setting" anything.

This is more in-line with typical controlled components, like forms, where you get an `onChange` prop callback and are responsible for updating your state to present the updated prop back to the controlled component

packages/react-heartwood-components/src/components/View/View.js
```diff
@@ -11,79 +11,60 @@ type Props = {
...
+		setSidebarVisibility,
```
gcpantazis: Same here

packages/react-heartwood-components/src/components/Core/components/HeaderPrimary/HeaderPrimary.js
```diff
@@ -135,7 +135,7 @@ export default class HeaderPrimary extends Component<Props, State> {
 							)}
 							<UserMenu
 								menuIsVisible={isUserMenuVisible}
-								toggleMenu={this.toggleUserMenuVisibility}
+								setMenu={this.setUserMenuVisibility}
```
gcpantazis: Did you add this prop to `UserMenu` or was it already there?

CarolynSM: I didn't add it but I did change the name....should this one have been left as `toggle`?

gcpantazis: I just didn't see the code here where you made that change. This is fine if that's checked in already.

PR #260, opened 2018-12-19
GraphQL, Node 10, ACLs, Skill Views, Flow

kengoldfarb:
### MAJOR UPGRADE

With all the updates to functionality, this will increment the major version number to `v9`

* Skill Views

  * Start stubbing out support for responding w/ skill views for `location_dashboard`

* Flow now supported in `spruce-skill` (still need to add types a lot of places though)

* GraphQL support

  * Koa middleware to expose GQL if `DB_ENABLED=true` and `GRAPHQL_ENABLED=true`

  * Helpers to automate some of the tedious setup and handle scoping

  * `spruce-node` query and mutation wrappers for making GQL calls against the api

  * Queries

    * Users

    * UserLocations

* ACLS

  * Skills register acls that they publish and those they wish to receive

  * `props.auth.acl` in `spruce-skill` is automatically populated with skill's requested ACLs for the current user

  * `ctx.services.acl` implemented in `spruce-skill-server` with methods to fetch ACLs from the api

  * GraphQL resolvers can access the current user on `context.auth.User`

* Upgrade to node 10

* Minor stuff

  * Kill debugger on skill restart to prevent error about address already in use

  * Change default log level to `info` (previously `warn`)

## Type

- [X] Feature
- [ ] Bug
- [X] Tech debt

### DIFF
--- a/packages/spruce-skill-server/gql/helpers.js
+++ b/packages/spruce-skill-server/gql/helpers.js
@@ -1,4 +1,3 @@
-// @flow
 const parseFields = require('graphql-parse-fields')
 // Sequelize uses the inflection library for it's naming
 const inflection = require('inflection')
@@ -16,7 +15,7 @@ const {
 	defaultArgs
 } = require('graphql-sequelize')
 
-function pathToScope(path: Object, prevPath?: string) {
+function pathToScope(path, prevPath) {
 	let scope = ''
 	if (prevPath && typeof path.key === 'string') {
 		scope += `${path.key}.${prevPath}`
@@ -31,7 +30,7 @@ function pathToScope(path: Object, prevPath?: string) {
 	return scope
 }
 
-function enhancedAttributeFields(model: any, options?: Object) {
+function enhancedAttributeFields(model, options) {
 	if (!options) {
 		options = {}
 	}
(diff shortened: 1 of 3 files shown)


packages/spruce-skill-server/factories/sequelize.js
```diff
@@ -41,10 +41,12 @@ module.exports = (
 		log.info('Metrics: Sequelize hooks disabled')
 	}
 
-	const coreModels = fs
-		.readdirSync(defaultModelsDir)
-		.filter(filterFile)
-		.map(file => path.resolve(defaultModelsDir, file))
+	const coreModels = fs.existsSync(defaultModelsDir)
+		? fs
```
liquidg3: is this for legacy? Are shipping with User, Location, and Organization models?

liquidg3: Also, is this wrapped higher up in a DB_ENABLED=true check?

packages/spruce-skill-server/services/acl.js
```diff
@@ -0,0 +1,91 @@
...
+	return `{${props}}`
+}
+
+module.exports = {
+	async getAcls(options) {
+		const { userId, permissions, locationId, organizationId } = options
+		const query = `
+		{
+			Acls (
+				userId: "${userId}"
```
liquidg3: do these need to be template literals? \`${userId}\`

PR #266, opened 2018-12-20
feature(tabs): Defer rendering of anchors to downstream code

gcpantazis:
- This is a bit of a proto for how I want to handle anchor rendering 
generally in react-heartwood; I often want to control anchor rendering 
so that I can instrument client side routing with `next/link`, etc.

- [x] Feature
- [ ] Bug
- [ ] Tech debt


gcpantazis:
I just remembered that `Button` handles this. Needs a little more work.

PR #270, opened 2018-12-24
Tabs — Ability to Truncate

jtini:
[SBL-1599](https://sprucelabsai.atlassian.net/browse/SBL-1599)

## Description
tabs

This makes any instance of Tabs automatically truncate itself. A couple of notes:
- To avoid this behavior (as needed for Onboarding Cards) you can pass `isTruncatable={false}`
- The window resize event is debounced by 500ms for performance. We could make this faster if it seems too slow.
- If the active tab would be hidden, it will instead be shown first, with the rest of the tabs in the order in which they were authored. The active tab will never appear in the context menu.
- The Tabs component uses state in an atypical fashion; namely that `tabWidths` and `contextTabWidth` are sort of permanent. These values are captured in the initial render so that we can easily reason about when to hide and show which tabs on window resize without having to render an ugly tab group.

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt

### DIFF
--- a/packages/heartwood-components/components/07-navigation/01-tabs/tabs.config.js
+++ b/packages/heartwood-components/components/07-navigation/01-tabs/tabs.config.js
@@ -1,8 +1,8 @@
 module.exports = {
 	title: 'Tabs',
 	collated: true,
-	collator: function(markup, item) {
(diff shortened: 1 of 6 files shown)


jtini:
Sorry for the crappy GIF — I promise it's more performant than that makes it look!

PR #289, opened 2019-01-05
Update canary with dev version

gcpantazis:
Might want to automate this. Self-merging for now.


gcpantazis:
Yikes, doing this on GH has some issues. It probably makes sense to merge `dev` into `canary` after a successful `dev` deploy, as a part of CI. That way the canary version is reflective of what will happen on the next canary->dev merge. 💫😵💫

PR #295, opened 2019-01-10
<List> Updates

CarolynSM:
## Description

⚡️Update`<List>` component to accept `items` as an optional prop and render children when passed

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt


packages/react-heartwood-components/src/components/List/components/PermissionsListItem/PermissionsListItem.js
```diff
@@ -0,0 +1,77 @@
+// @flow
```
liquidg3: I don't think this should be in workspace since no other skills are going to have access. instead, move to web.

thoughts @gcpantazis @jtini

jtini: I see what you mean, but if we keep it here, we can include it in Storybook.

CarolynSM: @liquidg3 updated to incorporate only the updates to `<List>`

liquidg3: this is good, ready to review web ai when it's ready @CarolynSM

PR #308, opened 2019-01-15
fix: Don't show the disabled error message

gcpantazis:
The other errors are fine, since they're predicated on wsUri being set. But this one logs every time a new client is created / there's no `window`, which happens a lot in dev on NextJS SSR.

## Type

- [ ] Feature
- [x] Bug
- [ ] Tech debt


kengoldfarb:
Maybe we change it to a `trace` log...this log will be useful for debugging GQL subscriptions

gcpantazis:
This isn't really as bothersome as I thought at first, I'll revisit with a better log-injection method later.

PR #321, opened 2019-01-17
Org header

brandonminch:
## Description

Updated primary header to allow organization with image

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt


packages/heartwood-components/components/99-web/header-primary.scss
```diff
@@ -90,6 +91,18 @@
 	}
 }
 
+.header-primary__organization-image {
+	background-position: center;
+	background-size: cover;
```
jtini: This is gonna crop anything that isn't square. Might not be an issue, but we would have to keep that in mind for whichever step the Owner adds this logo image.

PR #322, opened 2019-01-17
feat(eslint): Add spruce eslint plugin

gcpantazis:
- Adding this to aid in GQL rules enforcement; easier to apply a rule and get all the errors in the project than manually looking around for them. Plus it future-proofs the code.
- Going forward I'd much prefer to set a standard, write a linter rule to enforce it, then hand-off to one/many engineers to implement.
- This is a medium issue right now where we have several branches with mixed standards in ai-spruce-web; It's hard to refactor when new code might come in and break the new rules. Having a linter makes  this easy, I just tell everyone to merge up dev and they'll have the new parameters flagged in their IDEs (or otherwise they'll see them on CircleCI when they open a PR).
- I can move it somewhere else if we don't want it in workspace, but I dig the deploy process here and all 🤷‍♂️ LMK what you think on that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
 	"private": true,
 	"name": "workspace.sprucebot-skills-kit",
 	"workspaces": [
+		"packages/eslint-plugin-spruce",
 		"packages/heartwood-components",
 		"packages/react-heartwood-components",
 		"packages/spruce-next-helpers",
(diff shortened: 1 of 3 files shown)


gcpantazis:
Regex had a ton of holes, just swapped it out with a bracket-counter to determine nesting 😂

PR #350, opened 2019-01-23
Toast

jtini:
[SBL-1866](https://sprucelabsai.atlassian.net/browse/SBL-1866)

## Description

Fixes for the `Toast` and `ToastWrapper` components:
- Updates styles to match latest from Heartwood
- Gives the `ToastWrapper` a fixed position so that it stays put on scroll
- Adds a `z-index` of 9 to keep Toasts above the UI
- Adds a default timeout of 3 seconds, with the option to override or remove timeout entirely
- Adds `id`s to Toasts in the ToastWrapper — this achieves timeout management and makes the animations smoother because we no longer rely on `idx` for updates
- Adds a `max-width` to avoid UI blocking

## Type

- [ ] Feature
- [ ] Bug
- [ *] Tech debt

### DIFF
--- a/packages/react-heartwood-components/src/components/Toast/components/ToastWrapper/ToastWrapper.js
+++ b/packages/react-heartwood-components/src/components/Toast/components/ToastWrapper/ToastWrapper.js
@@ -52,7 +52,7 @@ export default class ToastWrapper extends Component<Props, State> {
 	}
 
 	render() {
-		const { toasts } = this.props
+		const { toasts } = this.state
 
 		return (
 			<div className="toasts-wrapper">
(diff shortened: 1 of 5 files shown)


packages/react-heartwood-components/src/components/Toast/components/ToastWrapper/ToastWrapper.js
```diff
@@ -12,23 +13,63 @@ type Props = {
...
+			if (toast.timeout !== 'never') {
+				timeouts[toast.id] = setTimeout(
+					() => {
+						props.handleRemove(toast.id)
+					},
+					typeof toast.timeout === 'number' ? toast.timeout : defaultTimeout
+				)
+			}
+		})
+		return { toasts: uniqToasts, timeouts }
```
gcpantazis: I don't see where you're accessing `state.toasts`; is that necessary, or do you mean to access the newly unique toasts in `render()`?

jtini:
@brandonminch I agree. I fixed the color and changed it to `followupAction` and added `followupText` to keep things open-ended.

packages/react-heartwood-components/src/components/Toast/components/ToastWrapper/ToastWrapper.js
jtini: Good call — they should be accessed in `render()`. We need it here to set the right timeouts.

PR #354, opened 2019-01-24
Add ProfileSummary component

jtini:
## Description

Adds the `ProfileSummary` component for use in Guest and Teammate profiles. This component is only visible on small screens and its job is to provide a way for users to access the right sidebar.

## Type

- [X ] Feature
- [ ] Bug
- [ ] Tech debt


packages/heartwood-components/components/99-web/profile-summary.scss
```diff
@@ -0,0 +1,19 @@
+.profile-summary {
+	display: flex;
+	justify-content: space-between;
+	align-items: center;
+	width: 100%;
+	padding: rem(12) spacing('base');
+	background-color: $c-grey-00;
+	border-bottom: 1px solid $c-border-light;
+
+	@include gt('990px') {
```
gcpantazis: Why do this here vs. at the point where it's implemented (in a specific page, etc)

jtini: This behavior is gonna be shared between Guest and Teammate profile, although I suppose we could do the media query in a shared page layout. I'm cool with removing it here though, it does seem kinda brittle.

PR #355, opened 2019-01-24
Fix running skills in non-local environments

kengoldfarb:
## Description

Currently deploying in non-local environments will crash skills.

This changes which directory to run from based on `process.env.ENV`. If in a non-local environment we run from the `build/` directory which has had flowtypes stripped.

## Type

- [ ] Feature
- [X] Bug
- [ ] Tech debt

### DIFF
--- a/packages/spruce-skill/config/default.js
+++ b/packages/spruce-skill/config/default.js
@@ -17,7 +17,7 @@ try {
 
 // When running locally we use 'flow-node' so it can handle flowtypes. When in a non-local environment we need to use the build/ directory where flowtypes have been stripped
 const baseDirectory =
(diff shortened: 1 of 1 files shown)


packages/spruce-skill/config/default.js
```diff
@@ -15,6 +15,12 @@ try {
 	console.error('Missing .env file for this project')
 }
 
+// When running locally we use 'flow-node' so it can handle flowtypes. When in a non-local environment we need to use the build/ directory where flowtypes have been stripped
```
gcpantazis: Is there an advantage to this over using `@babel/flow`?

PR #359, opened 2019-01-24
fix: Add .editorconfigs targetting all files

gcpantazis:
- This ensures that non-JS files are formatted properly (JSON were 
getting space-tabbed in VSCode)
- Other than spruce-skill (which acts as a template) inherit rules from 
lerna root


.editorconfig
```diff
@@ -4,4 +4,3 @@ root = true
 trim_trailing_whitespace = false
 insert_final_newline = true
 indent_style = tab
-indent_size = 2
```
kengoldfarb: 👍

PR #364, opened 2019-01-28
Add Feed Builder

jtini:
[DES-255](https://sprucelabsai.atlassian.net/browse/DES-255)

## Description

Use the Feed Builder to render feeds of messages. This uses [React Virtualized](https://bvaughn.github.io/react-virtualized/#/components/List) to handle loading new messages on scroll and to handle thousands of messages that are already loaded.

Notes:
- To see this in action, run Storybook and visit the Feed Builder story (which fakes API calls to load new messages), or the Guest Profile story under View (which doesn't fake loading but shows the component in context)
- This component assumes that messages can be loaded in reverse-chronological order
- The default rule for `from` images is to hide them if the messages were sent within 30 minutes of each other, but seeing it in action, we may want to revisit this threshold
- There may be a small amount of scroll-jank with new messages. I didn't go too far with fixing this in the interest of time, and knowing that the component will probably behave differently when it's actually hitting the API

TODO:
- Add a loader for better feedback
- TBD on empty state; there is a prop for it, but we can also determine boilerplate messages if preferable

## Type

- [X ] Feature
- [ ] Bug
- [ ] Tech debt


packages/heartwood-components/components/feed/feed.scss
```diff
@@ -0,0 +1,69 @@
...
+// }
+
+// .message-feed {
+// 	height: calc(100vh - #{rem(48)});
+// }
+
+// .message-feed,
+// .message-feed__message-wrapper {
+// 	transform: scaleY(-1);
+// }
```
brandonminch: @jtini can the commented code be removed?

jtini: yep - will remove

packages/react-heartwood-components/src/components/Message/Message.js
```diff
@@ -81,53 +98,75 @@ export const Message = (props: MessageProps) => {
...
-			<span class="message__pre">
+	const renderImage = ({ image, alt, isFromSprucebot }) => {
+		if (isFromSprucebot) {
+			return (
 				<img
-					class="message__from-image"
-					src={fromImage || ''}
-					alt={fromAlt || fromName || ''}
+					className="message__from-image message__from-sprucebot-image"
+					src={image || ''}
```
gcpantazis: Probably return `null` if you don't have a source, this would just render a busted image

packages/react-heartwood-components/src/components/Message/Message-story.js
```diff
@@ -106,7 +111,14 @@ stories.addDecorator(story => (
 stories.addDecorator(withKnobs)
 
 stories
-	.add('Message', () => <Message>It's lonely over here.</Message>)
+	.add('Message', () => (
+		<Message
+			from={{ name: '', image: '', alt: '', id: '' }}
+			dateSent={moment()}
+		>
+			It's lonely over here.
```
gcpantazis: Encode `'` or wrap string in template literal

packages/react-heartwood-components/src/components/Core/components/FeedBuilder/FeedBuilder.js
```diff
@@ -0,0 +1,251 @@
...
+	id: string,
+	...MessageBuilderProps
+}
+
+type Props = {
+	/** Messages for the feed */
+	messages?: Array<MessageProps>,
+
+	/** The total number of messages in the feed. Used for infinite scrolling. */
+	messageCount?: number,
```
gcpantazis: Is this necessary? For infinite loading, wouldn't you just stop loading if you went back far enough and time and got zero results? I actually thought that was the default behavior of the virtualized loader component

jtini: @gcpantazis I added this to make scrolling up smoother when new items are added. It sets the `scrollToIndex` to keep the position consistent when new items load in. This also prevents the `loadMoreRows` callback from firing over and over again by scrolling down.

gcpantazis: I feel like this wasn't necessary the last time I made one of these. And plus I don't know that we can rely on being able to figure out how many thousands of items might be included in some feeds... and chats get to the thousands of entries very very quickly.

gcpantazis: @jtini I took a look at my last project that used this, and it seemed like I handled this by `return`ing out of `loadMoreRows` if I was already loading something (which in that case I tracked by setting `isLoading` to true on state). This worked really nicely and let me load infinitely backwards without knowing for certain how far back the history went

PR #365, opened 2019-01-28
Tabs — Fix wrap

jtini:
## Description

Addresses the wrapping issue with tabs — tabs will no longer break lines and through off their container's height.

## Type

- [ ] Feature
- [X ] Bug
- [ ] Tech debt


packages/heartwood-components/components/07-navigation/tabs.scss
```diff
@@ -55,8 +55,8 @@
 
 	&.tab-group--spacing-even {
 		.tab {
-			flex: 1 1 auto;
-			padding: 0;
+			flex: 1 0 auto;
+			// padding: 0;
```
brandonminch: @jtini can this line be removed? Otherwise everything else looks good!

PR #367, opened 2019-01-29
form builder and settings

liquidg3:
Form builder now works hand in hand with skills to render settings forms, load their values, and save them, along with skill side validation on field blur.

- [x] Feature
- [ ] Bug
- [ ] Tech debt


packages/spruce-next-helpers/src/skillskit/index.js
```diff
@@ -106,6 +30,9 @@ const skill = {
 	},
 
 	ready: function({ resetUrlTrail = false, showHeader = true } = {}) {
+		Iframes.sendMessage({
```
gcpantazis: 🤔 did you mean to leave this here? Not sending anything.

packages/spruce-skill/server/events/validate-settings.js
```diff
@@ -3,41 +3,45 @@ const { eventError } = require('../lib/errorHandler')
...
 	try {
 		log.debug('**** validate-settings', { event: ctx.event })
-		if (
-			!ctx.event.payload ||
-			!ctx.event.payload.settings ||
-			!ctx.event.payload.userId
-		) {
+		const { auth, event } = ctx
+
+		if (!auth.User || !auth.Organization) {
```
gcpantazis: Is this file only used for org-level settings?

packages/spruce-next-helpers/src/skillskit/index.js
liquidg3: oops, yeah, did not mean to leave that there!

packages/spruce-skill/server/events/validate-settings.js
liquidg3: nope, added a location check there.

PR #369, opened 2019-02-01
save bar and iframe messaging

liquidg3:
duration disabled
duration fully controlled
duration onChange if selecting value 
iframe messaging
autosuggest disabling
removed unnecessary methods from spruce-next-helpers/skillskit/index
moved skillskit/index to new iframe messaging util
save bar functionality
removed unused listeners from PageWrapper
removed checks on domain from iframe messenger


packages/react-heartwood-components/src/components/Forms/components/DurationInput/DurationInput.js
```diff
@@ -45,13 +52,41 @@ export default class DurationInput extends Component<Props, State> {
...
+		if (props.value !== state.prevPropsValue) {
+			return {
+				prevPropsValue: props.value,
+				value: DurationInput.minutesToStr(props.value)
+			}
+		}
+		return null
+	}
+
+	static minutesToStr = (num: number): string => {
```
gcpantazis: Note, but output types in flow are usually inferred. you don't need to include them unless you want to force the output to a type (useful sometimes but not always). In this case I think you're fine to omit it.

packages/spruce-utils/lib/iframes.js
```diff
@@ -41,39 +41,49 @@ export default class Iframes {
 
 	static onMessage(eventName: string, callback: Function) {
 		const responseHandler = (event: Object) => {
-			const RootDomainRegex = new RegExp(/[\w\d-]+\.[\w\d-]+$/)
-			const thisRootDomain = window.location.hostname.match(RootDomainRegex)
-			const sourceRootDomain = event.source.location.hostname.match(
-				RootDomainRegex
-			)
+			// TODO George skills will be hosted on external domains, so we may need to pass
```
gcpantazis: 🙆‍♂️

packages/spruce-utils/lib/iframes.js
```diff
@@ -41,39 +41,49 @@ export default class Iframes {
...
 
-					callback(event.data.dataFromUser, responder)
-				}
+				// legacy support
+				const data =
+					typeof event.data === 'string' ? eventData : event.data.dataFromUser
+
+				callback(data, responder)
 			}
+			// }
```
gcpantazis: Remove

PR #371, opened 2019-02-01
chore(react): Get heartwood on shared eslint config

gcpantazis:
- Biggest change is restricting unused vars, but this also caught several other issues with the code. Please flag anything I removed that should be left in commented out.

- [ ] Feature
- [ ] Bug
- [x] Tech debt


packages/react-heartwood-components/src/components/Button/Button-story.js
```diff
@@ -30,7 +24,6 @@ stories.addDecorator(withKnobs)
 stories
 	.add('Primary', () => (
 		<Button
-			isSmall={isSmall}
```
CarolynSM: don't know if we should keep this prop in and replace as `isSmall={boolean('isSmall', false}` so it appears with the knobs in storybook? that might be why it was written this way. @jtini ?

packages/react-heartwood-components/src/components/Core/components/SaveBar/SaveBar-story.js
```diff
@@ -45,7 +34,6 @@ stories
 			isDiscardDisabled={boolean('isDiscardDisabled', false)}
 			onDiscard={() => console.log('DISCARD')}
 			onSave={() => console.log('SAVE')}
-			isSaving={true}
```
CarolynSM: same with the button story, is there a better way to include this such that it'll come up in the knobs?

packages/react-heartwood-components/src/components/BigCalendar/components/DragGrid/DragGrid.js
```diff
@@ -966,34 +965,11 @@ class DragGrid extends PureComponent<Props, State> {
 		const {
 			className,
 			children,
-			onScroll,
```
liquidg3: @gcpantazis these props were not supposed to be dropped into the `...props` spread so it would not be passed to child components and I don't see them being deleted from props after. I can't promise there won't be side effects here.

gcpantazis: Ugh. IMO the `...props` spread pattern is dangerous, I get the convenience but it causes a lot of unintended consequences. The reverse is true too, a benign prop that we add higher up might echo into components and cause side-effects. I think it's better to lean into flow and be fairly explicit with our types.

Here though... I'll take a gander and see what impact the props spread will cause. If I see anything particularly concerning I'll fix.

gcpantazis: @liquidg3 I took a look at how this spread was applied in code, and found that the only inbound prop that was being spread was `style`. So I got rid of the spread, killed the dead vars, and just passed the `style` prop to the root div directly. LMK if you have any concerns with that 🙌

packages/react-heartwood-components/src/components/Button/Button-story.js
```diff
@@ -88,7 +79,6 @@ stories
 	))
 	.add('Caution', () => (
 		<Button
-			isSmall={isSmall}
```
brandonminch: @gcpantazis curious why these got removed. looks like it was using the boolean knob to allow `isSmall` to be toggled on or off in storybook.

brandonminch: nvm, saw the resolved comment from carolyn above!

PR #380, opened 2019-02-06
better examples

liquidg3:
make sure example views and cards are everywhere


packages/spruce-skill/interface/pages/skill-views/public_location/index.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import Router from 'next/router'
+// import Router from 'next/router'
```
gcpantazis: Can we remove all these commented-out Router imports?

liquidg3: I’m trying to balance our linting  with wanting to show examples.

Where we are today is the example pages add almost no value since they are so bare.

What do you think?

gcpantazis: I get wanting to show examples, and think we should include (commented out) things that the skills dev needs to get their job done that are particular to Spruce. But I'd probably draw the line at something external to us (a NextJS lib in this case) since I think the skills dev should be aware that they're using Next and read Next's documentation on what's possible there.

PR #381, opened 2019-02-06
fix(flow): React Heartwood, correct corrected flow config

gcpantazis:
Some of my work yesterday was based on issues I was having in Atom; this turned out to be related to https://nuclide.io/ being retired by Facebook (oops) and some of the packages therein turning to mush.

So anyway, now I'm full steam ahead on the good ship VSCode... this commit should finish the previous PR's work and get flow type-checking correct in react-heartwood. Still 1k+ errors but they're the *right* errors! 🎉


gcpantazis:
cc/ @kengoldfarb for when you get back ☝️ I went ahead and unified the flow configs for workspace; it just didn't work well to have them separated, IDEs can't handle multiple flow instances. You could have a per-project flow config but you could only run them in IDEs if you had the packages open in separate windows. And this way we'll all feel the communal pain of ~800 flow issues 😅

PR #383, opened 2019-02-06
Style icons used in text component and text links

brandonminch:
## Description

Style icons inside of `<Text>` components

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt

### DIFF
--- a/packages/heartwood-components/components/10-text/text.scss
+++ b/packages/heartwood-components/components/10-text/text.scss
@@ -38,6 +38,11 @@
 	font-weight: $font-weight-normal;
 	font-feature-settings: 'onum', 'pnum';
 	font-variant-numeric: oldstyle-nums proportional-nums;
+	.icon {
+		height: 1rem;
+		margin: 0 0.2rem;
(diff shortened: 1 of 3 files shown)


packages/heartwood-components/components/10-text/text.scss
```diff
@@ -38,6 +38,28 @@
...
+	display: inline-flex;
+	align-items: center;
+	.icon {
+		&:not(.u-icon__no-fill) {
+			fill: $c-primary;
+		}
+		&.u-icon__stroke {
+			stroke: $c-primary;
+		}
+		height: 1rem;
```
jtini: @brandonminch I've been trying to size these to match the x-height of the type they accompany. I don't know the exact value for Source Sans Pro, but I think that `0.85em` would be safe, even with other fonts. I do think we should size in `em` here since we want the icon's dimensions to be based on the text rather than the root.

brandonminch: sounds good. made the update

PR #396, opened 2019-02-13
feat(spruce-skill): Add example of requesting that Core show a toast

gcpantazis:
See https://github.com/sprucelabsai/ai-spruce-web/pull/144

core-toasts
toasts-from-skill

### DIFF
--- a/packages/react-heartwood-components/src/components/Toast/Toast.js
+++ b/packages/react-heartwood-components/src/components/Toast/Toast.js
@@ -23,6 +23,9 @@ const ToastHeader = (props: HeaderProps) => {
 }
 
 export type Props = {
+	/** Unique ID for the toast */
+	id: string,
+
 	/** Headline text */
 	headline: string,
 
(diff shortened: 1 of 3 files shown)


packages/react-heartwood-components/src/components/Toast/Toast.js
```diff
@@ -23,6 +23,9 @@ const ToastHeader = (props: HeaderProps) => {
 }
 
 export type Props = {
+	/** Unique ID for the toast */
+	id: string,
```
justinhillsjohnson: @gcpantazis Can this also be a number?

gcpantazis: 🔥thanks man, fixed

packages/spruce-skill/interface/pages/skill-views/example.js
```diff
@@ -52,6 +52,31 @@ class TestSkillView extends React.Component<Props> {
 										}
 									})
 								}}
+							/>{' '}
```
CarolynSM: what's this space here for? I think you'll want to wrap these buttons in a `<ButtonGroup>` to space the buttons appropriately

gcpantazis: Sure, done

PR #397, opened 2019-02-13
Paged Modal

brandonminch:
## Description

Create a Paged Modal component that allows paging through multi-stepped modals.

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt

### DIFF
--- a/packages/heartwood-components/components/06-modal/modal.scss
+++ b/packages/heartwood-components/components/06-modal/modal.scss
@@ -13,6 +13,19 @@
 	justify-content: space-between;
 	padding: spacing('base');
 	border-bottom: 1px solid $c-border;
+
+	&--is-paginated {
+		.modal-header__title-wrapper {
+			flex: 1;
+		}
+		.modal-header__title {
+			text-align: center;
+			width: 100%;
+		}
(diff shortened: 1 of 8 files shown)


packages/heartwood-components/components/06-modal/modal.scss
```diff
@@ -21,11 +21,14 @@
...
 .modal-header__title-wrapper {
 	display: inline-flex;
+	flex: 1;
 }
 
 .modal-header__title {
 	@include typescale('heading-lg');
 	margin-top: rem(2);
+	width: 100%;
+	text-align: center;
```
jtini: @brandonminch I don't think this centering is gonna work, because the text is not visually centered when there is no back button. This would make a lot of our existing modals (especially longer titled ones) look lopsided. Also, I don't think this should be the default for _all_ modals — I'd rather configure it with something like `isCentered` on `Modal.Header` and `PagedModal` could just pass that in by default.

brandonminch: whoops, yeah I meant to scope it to paged-modal but I'll make it a prop instead

PR #407, opened 2019-02-20
fix(Table): Correct busted logic that was breaking pagination

gcpantazis:
- Introduced in #379 to remove pagination when record count was below 
the limit but not tested for the opposite case.

## Type

- [ ] Feature
- [x] Bug
- [ ] Tech debt


justinhillsjohnson:
@gcpantazis This one looks ready to go!

PR #414, opened 2019-02-26
GQL Relay and other improvements

kengoldfarb:
[SBL-2005](https://sprucelabsai.atlassian.net/browse/SBL-2005)

## Description

* Fix v2 authentication to use core API w/ relay

* Implement GQL relay

* Code cleanup

* Add better logging and make startup less verbose

* Add new env variable: `API_ONLY=true` to have only the api boot. This is useful when testing just the BE so we don't have to wait for next.js to build

* Changed eslint `no-unused-vars` to warning. I'd like to be able to show the available variables in some places without actually using them...open to other suggestions though @gcpantazis 

* Re-enable `get-views` defaults that were commented out

## Type

- [X] Feature
- [ ] Bug
- [X] Tech debt


packages/eslint-config-spruce/index.js
```diff
@@ -15,7 +15,7 @@ module.exports = {
 		'no-undef': 'error',
 		'no-var': 'error',
 		'no-unreachable': 'error',
-		'no-unused-vars': 'error',
+		'no-unused-vars': 'warn',
```
gcpantazis: Can you just comment vars out if you don't actually want to use them? If this is for a specific project (like spruce-skill) where that is onerous you can just disable it per-repo in that project's eslintrc. I'd like the default to be harsher.

packages/spruce-skill-server/index.js
```diff
@@ -349,49 +364,51 @@ module.exports = async ({
...
-					body = `Redirecting to ${ctx.res.redirect}`
-					ctx.redirect(ctx.res.redirect)
-					ctx.res.end(body)
-					// return
+			debug('handing off to next and backing off', ctx.path, ctx)
+			await handle(ctx.req, ctx.res)
+			ctx.respond = false
+			return
+
+			// this does not work as desired
```
gcpantazis: Is this still true? 😬 

I know this was existing code and you just reformatted but could you add more details?

packages/eslint-config-spruce/index.js
gcpantazis: To expand on that, I see more dead vars than I do "example vars" in code. Enforcing this has been a benefit to the projects I added it for, since it 1) exposed references that had been forgotten about during our many refactors and 2) prevented people from introducing new unnecessary code during their active dev. 

I do think I see the benefit to skipping this rule in `spruce-skill`, but as a counterpoint, when I went through that repo to apply this rule I found that probably 50% of the unused vars weren't example code, they were refs that weren't doing anything in that file / weren't helpful to the end-user eng. The remaining 50% I typically commented out. I think this forces us to consider what is and isn't important as an example in these more "template"-like projects.

That said though, I'm not currently active in maintaining that project and if it feels onerous to maintain the commented-out variables I'm fine disabling it there. In other projects, I'd like to maintain the rule as-is.

packages/spruce-skill-server/gql/listeners.js
```diff
@@ -26,7 +27,7 @@ module.exports = (koa, gqlOptions) => {
 
 	listenerPaths.forEach(path => {
 		try {
-			log.info(`Importing gql listener: ${path}`)
+			debug(`Importing gql listener: ${path}`)
```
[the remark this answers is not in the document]
kengoldfarb: It won't print out every file it imports by default and will only report errors.

Unless `DEBUG=spruce-skill-server` is set, then it will be verbose.  This is the same pattern we follow elsewhere for verbose logs in packages

packages/spruce-skill-server/gql/router.js
```diff
@@ -136,10 +136,31 @@ module.exports = (koa, gqlOptions, server) => {
 					context
 				}) => {
 					const ms = log.timerEnd(context.startTime)
+
+					context.warnings = _.uniq(context.warnings)
+					context.attributeWarnings = _.uniq(context.attributeWarnings)
+					context.scopeInfo = _.uniq(context.scopeInfo)
+
+					if (context.warnings.length > 0) {
```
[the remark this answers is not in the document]
kengoldfarb: Yep, these checks are unnecessary

packages/spruce-skill-server/index.js
kengoldfarb: I didn't write this...not sure what this was referring to. @liquidg3 do you know?

packages/eslint-config-spruce/index.js
kengoldfarb: @gcpantazis Updated this to make it `error` again.

I have an override in `spruce-skill` that allows just `ctx` to be unused

PR #421, opened 2019-03-01
optional prop to hide count text in <RecordSelectList>

CarolynSM:
## Description

Add optional prop to `<RecordSelectionList>` to hide "# select" text

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt


justinhillsjohnson:
Looks good to me

PR #427, opened 2019-03-08
Prerelease/core modal footer

brandonminch:
## Description

* Adds modal footer action helper methods to skillskit
* Add example modal to spruce-skill.
* Adds `renderLocation` prop to pages that will add a `render-location-LOCATION_NAME` class to skill page `<body>` for styling based on where the page is being rendered. Current options are `page | modal | right-rail` and default is `page`
* Add ability to use a paged modal in core and update footer action and title text 

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt


packages/spruce-next-helpers/src/skillskit/next/_document.js
```diff
@@ -41,6 +43,12 @@ export default class MyDocument extends Document {
 				? `skill-${publicRuntimeConfig.SLUG}`
 				: ''
 
+		const renderLocation = get(
+			this.props.__NEXT_DATA__,
+			'props.pageProps.initialProps.renderLocation',
+			null
+		)
+
```
brandonminch: @gcpantazis this `__NEXT_DATA__` object was the only way I could find to access the `renderLocation` prop that is set on the page. Feels a little dirty but the idea here was so that the `render-location-xxx` class would get added even if the page is rendered server-side.

packages/spruce-skill/interface/pages/skill-views/dashboard_location/index.js
```diff
@@ -24,6 +25,8 @@ const EXAMPLE_SUBSCRIPTION = gql`
 	}
 `
 class DashboardLocationPage extends React.Component {
+	_modal = this.props.skill.modal()
```
kengoldfarb: I don't think we've been denoting private properties w/ a leading underscore elsewhere. Let's afterparty tomorrow on convention @brandonminch @gcpantazis

packages/spruce-next-helpers/src/skillskit/next/_document.js
gcpantazis: Yeah, you'd want to do it like this:

https://github.com/sprucelabsai/ai-spruce-web/blob/dev/pages/_document.js#L14-L18

This would spread initial props for the page into the _document render's props.

brandonminch:
@gcpantazis I didn't realize this had already been approved but I added to it the ability to show a paged modal in core. Let me know if you want me to open a separate PR!

PR #432, opened 2019-03-13
feat(view): Switch to a flexbox-based layout

gcpantazis:
- The old layout used a fixed-position layout w/ large paddings to 
account for sidebars
- Switching to flexbox so that the containers have appropriate natural 
widths, which is necessary for making components that "fit within their 
bounds" (such as calendar in core)
- Also fix some overflow bar issues for mouse users ([SBL-2240]) but not 
all.
- Update all examples in storybook
- One major change to `Page`: in order to get a proper container 
heirarchy, I've added a `sidebar` prop, which can be used to inject a 
sidebar element. This deprecates `hasSidebar`, since the layout no 
longer cares whether or not a sidebar is to the right of the content 
(flexbox distributes space instead)

## Type

- [x] Feature
- [ ] Bug
- [x] Tech debt

### DIFF
--- a/packages/react-heartwood-components/src/components/View/View-story.js
+++ b/packages/react-heartwood-components/src/components/View/View-story.js
@@ -236,57 +236,59 @@ class SkillViewExample extends Component<SkillViewProps, SkillViewState> {
 						},
 						tabs: skillViewTabs
 					}}
-					hasSidebar="large"
 					sidebarIsCollapsed={!sidebarsExpanded.right}
-				>
-					<Sidebar
-						side="right"
-						isCollapsible={false}
-						isLarge
-						isExpanded={sidebarsExpanded.right}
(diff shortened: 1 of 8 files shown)


packages/react-heartwood-components/src/components/View/View-story.js
```diff
@@ -236,57 +236,59 @@ class SkillViewExample extends Component<SkillViewProps, SkillViewState> {
...
+												}
+											]}
+										/>
+									</CardFooter>
+								</Card>
+							</SidebarSection>
+						</Sidebar>
+					}
+				>
+					<div style={{ flex: 1 }}>hey</div>
```
justinhillsjohnson: Debug code?

gcpantazis: Oops, yep. 🙇

PR #433, opened 2019-03-14
refactor(contextMenu): add max height and remove button rounded corners

brandonminch:
## Description

* Adds a max height so context menus with 5 or more actions will scroll
* Removes rounded corners for context menu buttons so hover state is flush with edges

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt


justinhillsjohnson:
Looks good to me

PR #438, opened 2019-03-18
feature(Loader): pass props through for useful things like additional…

justinhillsjohnson:
… styling

## Description

Perhaps this is too much of a direct pass-through of props, but let me know what you think. I want to be able to position the loader or style it with additional CSS properties based on how it's being used in Web.

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt

### DIFF
--- a/packages/react-heartwood-components/src/components/Loader/Loader.js
+++ b/packages/react-heartwood-components/src/components/Loader/Loader.js
@@ -11,13 +11,14 @@ type Props = {
 }
 
 const Loader = (props: Props) => {
-	const { isLight, isCentered } = props
+	const { isLight, isCentered, ...rest } = props
 	return (
 		<span
 			className={cx('loader', {
 				'loader--light': isLight,
 				'loader--centered': isCentered
 			})}
+			{...rest}
 		>
 			<span className="loader__dot" />
 			<span className="loader__dot" />


brandonminch:
@justinhillsjohnson is there a reason you'd want to pass styles to the loader vs targeting in a stylesheet on web for the specific use cases?

justinhillsjohnson:
> @justinhillsjohnson is there a reason you'd want to pass styles to the loader vs targeting in a stylesheet on web for the specific use cases?

@brandonminch Mainly because I *thought* I was keeping in line with other components, for example the `VHWrapper`: https://github.com/sprucelabsai/ai-spruce-web/blob/dev/components/VHWrapper/index.js + https://github.com/sprucelabsai/ai-spruce-web/blob/dev/components/IframePage/index.js#L28.

The real, immediate need is to differentiate between an inline loader and a full width/height loader (container). This can be done a bunch of ways so let me know what you might prefer!

brandonminch:
> > @justinhillsjohnson is there a reason you'd want to pass styles to the loader vs targeting in a stylesheet on web for the specific use cases?
> 
> @brandonminch Mainly because I _thought_ I was keeping in line with other components, for example the `VHWrapper`: https://github.com/sprucelabsai/ai-spruce-web/blob/dev/components/VHWrapper/index.js + https://github.com/sprucelabsai/ai-spruce-web/blob/dev/components/IframePage/index.js#L28.
> 
> The real, immediate need is to differentiate between an inline loader and a full width/height loader (container). This can be done a bunch of ways so let me know what you might prefer!

I think the only reservation I have with passing through props, in the case of heartwood components specifically, is that we give up a bit more control of the desired look or behavior of components. For your immediate use case would it make sense to set `display: inline-flex` by default and only make it `display: flex` if `isCentered` is true?

PR #439, opened 2019-03-19
v7 Hotfix: Allow deselecting date on DateSelect

brandonminch:
[SBL-2122](https://sprucelabsai.atlassian.net/browse/SBL-2122)

## Description

Adds the ability to deselect dates on the DateSelect component.

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt

### DIFF
--- a/packages/react-sprucebot/lib/components/Calendar/BigCalendar.js
+++ b/packages/react-sprucebot/lib/components/Calendar/BigCalendar.js
@@ -338,7 +338,7 @@ function (_Component) {
               };
 
               _this.setState({
-                optionsLoaded: (0, _toConsumableArray2.default)(optionsLoaded).concat([options]),
+                optionsLoaded: [].concat((0, _toConsumableArray2.default)(optionsLoaded), [options]),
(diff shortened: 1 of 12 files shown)


brandonminch:
@kengoldfarb, @gcpantazis and @ryanrd I made this update to the existing `v7` branch however it looks like V2 booking is on version `7.7.2` of `react-sprucebot` and the current release for v7 is `7.8.3`. I'm not sure what all of the changes were since then but I want to verify with you guys that I'm making this update in the right place and that we would want to include all of the updates since `7.7.2`?

Also it looks like the linter tests are failing for an unknown reason. They work locally in `react-sprucebot` but I had to push with `--no-verify` to bypass the pre-commit hook and now circleci is throwing another error...

Screen Shot 2019-03-19 at 9 55 43 AM

brandonminch:
FYI the only file that was actually changed was: `packages/react-sprucebot/src/components/DateSelect/DateSelect.js`. The rest of the changes were prettier fixes and generated files.

PR #440, opened 2019-03-19
fix: update yarn clean script with correct paths

brandonminch:
## Description

Update the `clean` script with correct paths to package node_module directories

## Type

- [ ] Feature
- [ ] Bug
- [X] Tech debt


justinhillsjohnson:
Good here

PR #444, opened 2019-03-22
fix(lint): Add config dep to our spruce eslint plugin

gcpantazis:
## Type

- [ ] Feature
- [X] Bug
- [ ] Tech debt


packages/eslint-config-spruce/package.json
```diff
@@ -18,20 +18,19 @@
 		"url": "https://github.com/sprucelabsai/workspace.sprucebot-skills-kit/issues"
 	},
 	"dependencies": {
```
gcpantazis: These changes are probably the most important in the PR; I'm making prettier and typescript peerDependencies, since we should be unified in which versions of that we're using. We can always `||` if we want to support multiples, but this would avoid lint mismatching prettier/TS in downstream projects.

Also using exact versions for everything (excepting the lerna colocated spruce plugin), which seems to have fixed the issues I was having with yarn nesting dependencies incorrectly. And in this context, exact is probably best; we want to make sure all code is linted in the exact same way.

justinhillsjohnson: > Also using exact versions for everything 

👍

PR #448, opened 2019-03-27
refactoring(NPM): Update deps for audit prep, peer deps

gcpantazis:
- Cleaning up *some* of the peer dependency warnings
- Cleaning up *some* of the audit issues. Can't do all, since a big problem right now is fractal (and that isn't in active development anymore).
- Removing a busted version of `cssstyle` that wasn't needed and was causing issues with typescript testing.

## Type

- [ ] Feature
- [ ] Bug
- [x] Tech debt


package.json
```diff
@@ -42,16 +42,19 @@
 		"heroku-postbuild": "yarn build"
 	},
 	"dependencies": {
```
gcpantazis: Note for this, if a child package sets a peer dependency, yarn expects the root package.json to include that dependency.

PR #451, opened 2019-03-29
Skills API v1 mocks / auth support for tests

kengoldfarb:
## Description

* Add ability to mock / intercept v1 api calls

* Intercepts the `ctx.sb.user` event during tests and properly fetches data from the mock DB which allows us to run tests locally against `/api/1.0/<role>` endpoints and simulate v1 events

* Fixes issue where incorrect env var (should be `DB_MIGRATIONS` instead of `DISABLE_MIGRATIONS`) was being set during tests

## Type

- [X] Feature
- [X] Bug
- [ ] Tech debt

### DIFF
--- a/packages/spruce-skill-server/tests/v1APIMocks.js
+++ b/packages/spruce-skill-server/tests/v1APIMocks.js
@@ -1,0 +1,36 @@
+const config = require('config')
+
+module.exports = ctx => ({
+	async get(path, query) {
+		const matches = path.match(/\/locations\/([^/]+)\/users\/([^/]+)/)
+		if (matches && matches[1] && matches[2]) {
+			// ctx.sb.user() has been called. Fetch the data from the DB. Used for v1 authentication
+			const locationId = matches[1]
+			const userId = matches[2]
+			const userLocation = await ctx.db.models.UserLocation.findOne({
+				where: {
+					UserId: userId,
+					LocationId: locationId
+				},
+				include: [
+					ctx.db.models.User,
+					{
+						model: ctx.db.models.Location,
+						include: [ctx.db.models.Organization]
+					}
+				]
+			})
+			return userLocation
+		}
+		return Promise.resolve({})
+	},
+	async post(path, data, query, method) {
+		return Promise.resolve({})
+	},
+	async patch(path, data, query) {
+		return Promise.resolve({})
+	},
+	async delete(path, query) {
+		return Promise.resolve({})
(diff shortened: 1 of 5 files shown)


packages/spruce-skill-server/tests/v1APIMocks.js
```diff
@@ -0,0 +1,36 @@
+const config = require('config')
+
+module.exports = ctx => ({
+	async get(path, query) {
+		const matches = path.match(/\/locations\/([^/]+)\/users\/([^/]+)/)
+		if (matches && matches[1] && matches[2]) {
+			// ctx.sb.user() has been called. Fetch the data from the DB. Used for v1 authentication
+			const locationId = matches[1]
```
justinhillsjohnson: Should we consider returning errors to the implementer by doing something like a `typeof` check on these id's? Or coerce with `parseInt()` or `parseFloat()` and check for `NaN`?

packages/spruce-skill/.vscode/launch.json
```diff
@@ -43,7 +43,7 @@
 				"ORM_LOGGING": "true",
 				"API_ONLY": "true",
 				"TESTING": "true",
-				"DISABLE_MIGRATIONS": "true"
+				"DB_MIGRATIONS": "false"
```
corbanb: @kengoldfarb did we need to change this? going to impact @ryanrd and the ulta team down stream I'd imagine.

kengoldfarb: `DISABLE_MIGRATIONS` has never been a real env var we're using. It's always been `DB_MIGRATIONS`

packages/spruce-skill-server/tests/v1APIMocks.js
kengoldfarb: I'm not really following what you mean. Ids in our system are UUIDs (strings). This part of the code is just intercepting the `ctx.sb.user()` call and looking up the user for use in our tests (so we don't rely on any external api calls for tests).

justinhillsjohnson: ahh, i thought it was for a mock server that we would be building requests against. additionally, good to know about the UUIDs. marking the conversation as resolved.

packages/spruce-skill/.vscode/launch.json
corbanb: sweet

PR #459, opened 2019-04-10
Pass core SandboxMock to custom mocks

kengoldfarb:
## What does this PR do?

* Passes the `SandboxMock` data into a skill's custom mocks so they can build data off the generated location, org, users, etc.

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt


packages/spruce-skill-server/tests/SpruceTest.js
```diff
@@ -56,7 +60,10 @@ module.exports = basePath => {
 						)
 					}
 					this.mocks[mock.key] = mock
-					await mock.setup(options)
+					await mock.setup({
+						...options,
+						sandbox
+					})
```
corbanb: @kengoldfarb can you provide an example in one of the tests on how this is used and extended?

PR #462, opened 2019-04-11
v1 event mock should fetch location

kengoldfarb:
## What does this PR do?

Hijacks the `ctx.sb.location()` call in tests and returns the proper location. Without this update `event.Location` is an empty object in events.

## Type

- [ ] Feature
- [ ] Bug
- [X] Tech debt

### DIFF
--- a/packages/spruce-skill-server/tests/v1APIMocks.js
+++ b/packages/spruce-skill-server/tests/v1APIMocks.js
@@ -2,7 +2,7 @@ const config = require('config')
 
 module.exports = ctx => ({
 	async get(path, query) {
-		let matches = path.match(/\/locations\/([^/]+)\/users\/([^/]+)/)
+		let matches = path.match(/\/locations\/([^/]+)\/users\/([^/]+)$/)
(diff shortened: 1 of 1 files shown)


packages/spruce-skill-server/tests/v1APIMocks.js
```diff
@@ -22,6 +22,20 @@ module.exports = ctx => ({
 			})
 			return userLocation
 		}
+
```
gcpantazis: Is this over-eager? Wouldn't it match deeper paths like `/locations/ID/something/ID`?

Though I suppose v1 is probably fairly locked down to new endpoints now, and this might not be concerning.

gcpantazis: Oops, off by a line. I'm talking about `path.match(/\/locations\/([^/]+)/)` on line 26.

kengoldfarb: Yeah...good point. I'll update the regex to be more specific

PR #468, opened 2019-04-22
Strict ACL checks & spruce-skill-server tests

kengoldfarb:
## What does this PR do?

* Adds strict ACL checks to the `ctx.services.acl. userIsAuthorizedForAcls()` method

  * Now assumes the user is *not* authorized and verifies that every ACL requested is valid

  * Now Throws errors on missing required parameters

* Adds check to API mock GQL server that the permission being checked is for the skill currently being tested

* Additional debug logging around ACLs and event contracts

* Removes (unused) jest and implements tests via mocha for `spruce-skill-server`

* Add tests around ACL check scenarios

* Update `package.json` scripts for tests to create the `tmp/` directory if it doesn't already exist

* Adds `this.ctx` to base test class for convenience

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt

- [SDEV3-852](https://sprucelabsai.atlassian.net/browse/SDEV3-852)

### DIFF
--- a/packages/spruce-skill-server/middleware/events.js
+++ b/packages/spruce-skill-server/middleware/events.js
@@ -14,6 +14,18 @@ module.exports = (router, options) => {
 		let body = ctx.request.body
 		const eventName = body && body.event
 		// setup if we are listening to this event
+		if (!config.eventContract || !config.eventContract.events) {
+			debug('No event contract found in config')
+		}
+
+		if (
+			config.eventContract &&
+			config.eventContract.events &&
+			!config.eventContract.events[eventName]
+		) {
+			debug(`No eventContract specified for: ${eventName}`)
+		}
+
 		if (
 			ctx.path === '/hook.json' &&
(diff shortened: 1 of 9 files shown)


justinhillsjohnson:
@corbanb @ryanrd @liquidg3 Can we get a review on this asap tomorrow?

packages/spruce-skill-server/middleware/events.js
```diff
@@ -14,6 +14,18 @@ module.exports = (router, options) => {
...
+		if (!config.eventContract || !config.eventContract.events) {
+			debug('No event contract found in config')
+		}
+
+		if (
+			config.eventContract &&
+			config.eventContract.events &&
+			!config.eventContract.events[eventName]
+		) {
+			debug(`No eventContract specified for: ${eventName}`)
```
ryanrd: I'm so excited this has been added =)

PR #470, opened 2019-04-23
feat(RecordSelectionList): Add reset method

gcpantazis:
- Since RSL maintains its own state for the sake of infinite scrolling, needed a method to reset the state.

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt

### DIFF
--- a/packages/react-heartwood-components/src/components/RecordSelectionList/RecordSelectionList.js
+++ b/packages/react-heartwood-components/src/components/RecordSelectionList/RecordSelectionList.js
@@ -103,11 +103,11 @@ export default class RecordSelectionList extends Component<
 	}
 
 	async componentDidMount() {
-		const { loadRecords } = this.props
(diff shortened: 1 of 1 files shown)


packages/react-heartwood-components/src/components/RecordSelectionList/RecordSelectionList.js
```diff
@@ -130,6 +130,23 @@ export default class RecordSelectionList extends Component<
 		}
 	}
 
+	async reset() {
+		const { loadRecords } = this.props
+
+		const initialRecords = await loadRecords({
+			offset: 0,
```
justinhillsjohnson: Should these values come from `defaultProps`?

gcpantazis: Yeah good call; I went ahead and did this across the file. Also inverted the default for another default prop, since that feature should be opt-in.

PR #480, opened 2019-05-15
Fix ctx.sb.mutation not properly calling api SDEV3-1353

kengoldfarb:
## What does this PR do?

* Fixes issue where the query was being incorrectly sent from the helper method.  It's now been simplified to re-use the `https adapter "query" method` since mutations/queries follow the same request format for GQL

* Add tests for `ctx.sb.query` and `ctx.sb.mutation` helper methods

## Type

- [ ] Feature
- [X] Bug
- [ ] Tech debt

- [SDEV3-1353](https://sprucelabsai.atlassian.net/browse/SDEV3-1353)

### DIFF
--- a/packages/spruce-skill-server/tests/GQLMethodTests.js
+++ b/packages/spruce-skill-server/tests/GQLMethodTests.js
@@ -8,7 +8,7 @@ class GQLMethodTests extends SpruceTest(`${__dirname}/../../spruce-skill/`) {
 		it('Can call query', () => this.doQuery())
 		it('Can not call query as mutation', () => this.doQueryAsMutate())
 		it('Can call mutation', () => this.doMutate())
-		it('Can call mutation as query', () => this.doMutateAsQuery())
+		it('Can not call mutation as query', () => this.doMutateAsQuery())
 	}
 
 	async doQuery() {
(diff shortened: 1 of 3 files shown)


packages/spruce-skill-server/tests/GQLMethodTests.js
```diff
@@ -0,0 +1,76 @@
...
+class GQLMethodTests extends SpruceTest(`${__dirname}/../../spruce-skill/`) {
+	setup() {
+		it('Can call query', () => this.doQuery())
+		it('Can not call query as mutation', () => this.doQueryAsMutate())
+		it('Can call mutation', () => this.doMutate())
+		it('Can not call mutation as query', () => this.doMutateAsQuery())
+	}
+
+	async doQuery() {
+		const result = await this.ctx.sb.query(`
```
gcpantazis: Use `graphql-tag` to tag these template literals and get consistent formatting. (Not sure if it's included by default, been a few weeks. If not don't worry about it)

kengoldfarb: @gcpantazis This method actually doesn't support using `graphql-tag` since it's making the api request directly (not using apollo client).

A future enhancement could be that we support both ways?  And/or we move to using the apollo client for backend gql api requests

PR #486, opened 2019-05-29
[SDEV3-1534] SDK is now exposing multiple orgs on the sandbox data

corbanb:
This will allow for more diverse tests during skills development and testing.

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt

- [SDEV3-1534](https://sprucelabsai.atlassian.net/browse/SDEV3-1534)

### DIFF
--- a/packages/spruce-skill-server/tests/mocks/SandboxMock.js
+++ b/packages/spruce-skill-server/tests/mocks/SandboxMock.js
@@ -33,10 +33,27 @@ module.exports = class SandboxMock {
 			numGuests: options.numGuests || 4
 		})
 
+		const sandboxOther = await this.createSandbox({
+			numLocations: options.numLocations || 1,
+			numOwners: options.numOwners || 3,
+			numGroupManagers: options.numGroupManagers || 3,
+			numManagers: options.numManagers || 2,
+			numTeammates: options.numTeammates || 3,
+			numGuests: options.numGuests || 4
+		})
+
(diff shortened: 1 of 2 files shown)


packages/spruce-skill-server/tests/mocks/SandboxMock.js
```diff
@@ -33,10 +33,27 @@ module.exports = class SandboxMock {
 			numGuests: options.numGuests || 4
 		})
 
+		const sandboxOther = await this.createSandbox({
```
kengoldfarb: It looks like `createSandbox()` is doing:
`this.skill = await this.createSkill()` on line 79

We should probably pull that out of that method or adjust it so we don't have an orphan `this.skill` that gets overwritten on the 2nd call.  Maybe it becomes `this.otherSkill`?

Also, in `SpruceTest.before()` let's set the other org/location for convenience...i.e.
```
this.otherOrganization = this.mocks.sandbox.otherOrganization
const otherLocationId = Object.keys(this.mocks.sandbox.otherLocations)[0]
this.otherLocation = this.mocks.sandbox.otherLocations[otherLocationId]
this.otherSkill = this.mocks.sandbox.otherSkill
```

corbanb: @kengoldfarb thinking we could move the skill to `sandbox.skill = await this.createSkill()` so the sandbox has its own skills as well? then we can grab different skills if we needed from the different sandboxes.

PR #495, opened 2019-06-07
[SDEV3-1593] Adds basic text templating to Text component

brandonminch:
## What does this PR do?

Adds ability to pass text templating syntax to Text component to enable rich formatting and interactions within text component

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt

- [SDEV3-1593](https://sprucelabsai.atlassian.net/browse/SDEV3-1593)

Screen Shot 2019-06-07 at 3 01 23 PM

text

### DIFF
--- a/packages/react-heartwood-components/src/components/Text/Text.js
+++ b/packages/react-heartwood-components/src/components/Text/Text.js
@@ -1,8 +1,58 @@
 // @flow
-import React from 'react'
+import React, { Fragment } from 'react'
 import cx from 'classnames'
 
+import Button from '../Button/Button'
+import TextStyle from '../TextStyle/TextStyle'
+
 import type { Node } from 'react'
+
+// Components available for templating
+
+const TextComponentKey = {
+	style: TextStyle,
+	button: Button
+}
+
+// Allows basic templating functionality on text strings
+
+const TemplateEngine = (text = '', context = {}) => {
+	let re = /{{([^}}]+)?}}/g,
+		children = [],
+		cursor = 0,
+		match
+
+	let add = function(line, js) {
+		if (line !== '') {
+			children.push({
+				props: { element: 'span', children: line.replace(/"/g, '\\"') }
+			})
+		}
+		if (context[js]) {
+			children.push(context[js])
+		}
+	}
+	while ((match = re.exec(text))) {
+		add(text.slice(cursor, match.index), match[1])
+		cursor = match.index + match[0].length
+	}
+	add(text.substr(cursor, text.length - cursor))
+
+	return children.map(renderText)
+}
+
+const renderText = child => {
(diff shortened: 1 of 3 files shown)


packages/react-heartwood-components/src/components/Text/Text.js
```diff
@@ -1,16 +1,69 @@
...
+		add(text.slice(cursor, match.index), match[1])
+		cursor = match.index + match[0].length
+	}
+	add(text.substr(cursor, text.length - cursor))
+
+	return children.map(renderText)
+}
+
+const renderText = child => {
+	console.log('CHILD', child)
```
gcpantazis: Remove

packages/react-heartwood-components/src/components/Text/Text.js
```diff
@@ -1,16 +1,69 @@
...
+
+// Allows basic templating functionality on text strings
+
+const TemplateEngine = (text = '', context = {}) => {
+	let re = /{{([^}}]+)?}}/g,
+		children = [],
+		cursor = 0,
+		match
+
+	let add = function(line, js) {
```
gcpantazis: js -> ?

`renderMethod` I think? Not sure what JS means here... other than JavaScript 😂

packages/react-heartwood-components/src/components/Text/Text.js
```diff
@@ -42,30 +95,36 @@ export type TextProps = {
 }
 
 const Text = (props: TextProps) => {
-	const { children, className, element, ...rest } = props
+	const { children, className, element, context, ...rest } = props
 	let Element = 'p'
+	let text = children
```
gcpantazis: `text` might not be text, like you see below; maybe store original children as `originalChildren`, then `let children = originalChildren` so you can override it. Makes it a little more compatible w/ the usual usages of `{children}`

packages/react-heartwood-components/src/components/Text/Text.js
```diff
@@ -1,16 +1,69 @@
...
+		if (line !== '') {
+			children.push({
+				props: { element: 'span', children: line.replace(/"/g, '\\"') }
+			})
+		}
+		if (context[js]) {
+			children.push(context[js])
+		}
+	}
+	while ((match = re.exec(text))) {
```
gcpantazis: Standard warning to be absolutely sure there's no way this can result in an endless loop 😬

gcpantazis: (Seems fine at a glance, though)

brandonminch: Updated to store matches first and then iterating over them while still calling exec in order to ensure a finite loop. Little weird but less scary...

PR #496, opened 2019-06-07
[SDEV3-934]: Add convenience methods for modifying button props

brandonminch:
## What does this PR do?

Adds convenience methods for modifying button props on confirm modals

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt

- [SDEV3-934](https://sprucelabsai.atlassian.net/browse/SDEV3-934)

buttons

### DIFF
--- a/packages/spruce-next-helpers/src/skillskit/index.js
+++ b/packages/spruce-next-helpers/src/skillskit/index.js
@@ -263,7 +263,15 @@ const skill = {
 
 	confirm: function() {
 		const confirm = {
-			show: (data: Object, onConfirm?: Function, onCancel?: Function) => {
+			show: ({
+				data,
+				onConfirm,
+				onCancel
+			}: {
+				data: Object,
+				onConfirm?: Function,
+				onCancel?: Function
+			}) => {
 				Iframes.sendMessage({
 					to: window.parent,
 					eventName: 'Confirm:Show',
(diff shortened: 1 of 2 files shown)


packages/spruce-skill/interface/pages/skill-views/dashboard_location/index.js
```diff
@@ -111,25 +111,25 @@ class DashboardLocationPage extends React.Component {
...
 			}
-		)
+		})
 	}
 
 	handleShowConfirmationWithInputModal = () => {
-		this.confirm.show(
-			{
+		this.confirm.show({
+			data: {
```
liquidg3: everywhere else i see us putting callbacks (`onConfirm` and `onCancel` on the same level as all other props. Check screen.

https://www.dropbox.com/s/tknfd6wofzrsfkh/Screenshot%202019-06-08%2012.01.03.png?dl=0

So i think we should drop `data` here and do it all on 1 level.

packages/spruce-next-helpers/src/skillskit/index.js
```diff
@@ -263,7 +263,15 @@ const skill = {
 
 	confirm: function() {
 		const confirm = {
-			show: (data: Object, onConfirm?: Function, onCancel?: Function) => {
+			show: ({
+				data,
```
liquidg3: See comment below, but we usually mixin callbacks with other props, like this screen: https://www.dropbox.com/s/ruzcb29dq6scstl/Screenshot%202019-06-08%2012.02.27.png?dl=0

so i think we should drop `data` from here.

packages/spruce-skill/interface/pages/skill-views/dashboard_location/index.js
brandonminch: updated

packages/spruce-next-helpers/src/skillskit/index.js
brandonminch: updated

PR #497, opened 2019-06-10
[SDEV3-1627] Record selector should have variable height

jtini:
## What does this PR do?

This addresses the issue of RecordSelectionLists and, by extension, RecordSelectors not having variable heights and never shrinking when they contain few items. This approach is a POC that you can see in Storybook, but isn't production ready, so please don't merge.

The idea here is to only use `react-virtualized` when the request returns a `totalRecordCount` large enough to require a virtualized list. Otherwise, a simple list is returned since all of the records can be fetched and displayed with a single request. This sidesteps the complications of shrinking `react-virtualized`'s `AutoSizer` component, which is designed only to grow, but not to shrink.

To make the determination, the component checks if `totalRecordCount` is greater than the length of `loadedRecords` after the first request. This may be an oversimplification as we might want to check for a specific number of records based on the context in which the component is used. In any case, this seemed like the most reliable and safest assumption to make.

## Type

- [ ] Feature
- [x] Bug
- [ ] Tech debt

- [SDEV3-1627](https://sprucelabsai.atlassian.net/browse/SDEV3-1627)

Nope

giphy

### DIFF
--- a/packages/heartwood-components/components/08-lists/lists.scss
+++ b/packages/heartwood-components/components/08-lists/lists.scss
@@ -238,7 +238,6 @@
 	}
 
 	.record-selection__list-wrapper {
-		// position: relative;
 		flex: 1;
 	}
 
--- a/packages/heartwood-components/components/08-lists/lists.scss
+++ b/packages/heartwood-components/components/08-lists/lists.scss
@@ -219,6 +219,16 @@
 }
 
 .record-selection__list {
+	display: flex;
+	flex-flow: column;
+	height: 100%;
+	overflow: hidden;
+
+	&:not(.record-selection__list--is-short) {
+		// Min Height is only needed for long lists
+		min-height: 28rem;
+	}
+
 	.text-container {
 		align-items: flex-start;
 		display: flex;
@@ -228,8 +238,8 @@
 	}
 
 	.record-selection__list-wrapper {
-		position: relative;
-		min-height: 28rem;
+		// position: relative;
+		flex: 1;
 	}
 
 	.record-selection__record-wrapper {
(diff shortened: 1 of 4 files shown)


packages/react-heartwood-components/src/components/RecordSelectionList/RecordSelectionList.js
```diff
@@ -310,60 +316,100 @@ export default class RecordSelectionList extends Component<
 				this.list.forceUpdateGrid()
 			}
 		}
+		const isListShort =
+			totalRecordCount &&
+			loadedRecords &&
+			loadedRecords.length === totalRecordCount
```
brandonminch: @jtini should this be `<=` in case records are filtered out for some reason but totalRecordCount remains?

packages/heartwood-components/components/08-lists/lists.scss
```diff
@@ -228,8 +238,8 @@
 	}
 
 	.record-selection__list-wrapper {
-		position: relative;
-		min-height: 28rem;
+		// position: relative;
```
brandonminch: can this be removed?

brandonminch:
👍 I think this solves it for all the cases that come to mind

packages/react-heartwood-components/src/components/RecordSelectionList/RecordSelectionList.js
jtini: @brandonminch Good point, but I don't think that can happen here. If I make that change, we'll get the non-virtual list 100% of the time. I think that `totalRecordCount` needs to be a reflection of the exact (filtered) request, not the actual total.

PR #499, opened 2019-06-10
[SDEV3-1590] Heartwood List Item: Entire Row Should be Clickable

jtini:
## What does this PR do?
Gives `ListItem` components the ability to be selectable either as checkboxes or radios. This also renders the list item text as `<label>`s so that clicking them toggles the checkbox or selects the radio option.

Adds the following props to `List` and `ListItem` respectively:
- `selectableType`: either 'checkbox' or 'radio'

- `checkboxId`
- `checkboxProps`

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt

- [SDEV3-1590](https://sprucelabsai.atlassian.net/browse/SDEV3-1590)

No

Screen Shot 2019-06-10 at 5 12 08 PM

### DIFF
--- a/packages/react-heartwood-components/src/components/List/components/ListItem/ListItem.js
+++ b/packages/react-heartwood-components/src/components/List/components/ListItem/ListItem.js
@@ -7,7 +7,7 @@ import Icon from '../../../Icon/Icon'
 import type { Props as ButtonProps } from '../../../Button/Button'
 import ContextMenu from '../../../ContextMenu/ContextMenu'
 import type { Props as ContextMenuProps } from '../../../ContextMenu/ContextMenu'
-import { Toggle } from '../../../Forms'
+import { Toggle, Checkbox, Radio } from '../../../Forms'
 import DragHandle from '../../../../../static/assets/icons/ic_drag_handle.svg'
 
 export type Props = {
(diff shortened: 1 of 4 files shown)


packages/react-heartwood-components/src/components/List/components/ListItem/ListItem.js
```diff
@@ -109,22 +121,50 @@ const ListItem = (props: Props) => {
...
+					{checkboxId && (
+						<Fragment>
+							{selectableType === 'checkbox' && (
+								<Checkbox id={checkboxId} {...checkboxProps} />
+							)}
+							{selectableType === 'radio' && (
+								<Radio id={checkboxId} {...checkboxProps} />
+							)}
+						</Fragment>
+					)}
```
brandonminch: @jtini `checkboxId` and `selectableType` are both optional but don't seem mutually exclusive. e.g. if I pass `selectableType` but not a `checkboxId` this wouldn't render my checkbox. Similarly I'm wondering if the `checkboxId` should be `selectableId` to eliminate confusion as to whether it's applies to both checkbox/radio

justinhillsjohnson:
Updated PR title to link the JIRA ticket to the PR (See https://cl.ly/a1a756d8fc67). Please use this style moving forward unless you have other recommendations. Thanks!

packages/react-heartwood-components/src/components/List/components/ListItem/ListItem.js
jtini: The non-mutual-exclusivity I think is fine, because it prevents id-less checkboxes/radios from being rendered that would cause problems. I agree about naming though, I'll change `checkboxId` and `checkboxProps` to `selectableId` and `selectableProps` respectively.

brandonminch: 👍

PR #510, opened 2019-06-14
[SDEV3-1716] Prerelease/record selection list

jtini:
## What does this PR do?

Makes the `RecordSelectionList` use `ListItem`s consistently. Previously, it was using a hacky workaround because the `ListItem` didn't support select-ability. Now that it does, this ensures that the right component is in place.

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt

None logged yet
- [SDEV3-1716](https://sprucelabsai.atlassian.net/browse/SDEV3-1716)

No, but it does deprecate the `renderRecord` prop and replaces it with `recordKeys` that tell `RecordSelectionList` how to map data into the `ListItem`s. This should ease dev effort as it will no longer be required to write out the render method for list items and instead you can just pass in the relevant keys.

This will also require a re-implementation effort eventually. I didn't remove the `renderRecord` prop, but without `recordKeys`, it won't render properly.

### DIFF
--- a/packages/react-heartwood-components/src/components/RecordSelectionList/RecordSelectionList.js
+++ b/packages/react-heartwood-components/src/components/RecordSelectionList/RecordSelectionList.js
@@ -14,6 +14,7 @@ import { TextInput, Checkbox, Radio } from '../Forms'
 import Button from '../Button/Button'
 import TextContainer from '../TextContainer/TextContainer'
 import Text from '../Text/Text'
+import ListItem from '../List/components/ListItem/ListItem'
 
 import type { Node } from 'react'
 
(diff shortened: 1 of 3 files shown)


packages/react-heartwood-components/src/components/RecordSelectionList/RecordSelectionList.js
```diff
@@ -245,69 +248,78 @@ export default class RecordSelectionList extends Component<
...
-							kind="simple"
-							className="record-selection__record-remove-btn"
-							disabled={false}
-							isSmall
-							icon={{ name: 'cancel_solid', className: 'btn__line-icon' }}
-							onClick={() => {
-								this.setState(
+					id={record.node[recordKeys.id]}
+					title={record.node[recordKeys.title]}
+					subtitle={record.node[recordKeys.subtitle]}
```
brandonminch: @jtini records will not always be in the format of `{ node: { ... } }` so we'll need a more flexible way to access the record keys now that the record isn't rendered inline with the component. I'd suggest we use lodash `get` on these so that recordKeys don't have to map directly to a property of the record. e.g. `title={get(record, recordKeys.title, '')}` would allow setting the `title` recordKey to `title` or `someChild.title`

PR #513, opened 2019-06-17
[SDEV3-1718] Autosuggest should not push input up or clip suggestions

jtini:
## What does this PR do?

Enables `Autosuggest` to use a [portal](https://reactjs.org/docs/portals.html) so that suggestions will not be clipped. Also, adds styling for the selected (via keyboard arrows) item. Verifiable in Storybook under Forms > Autosuggest.

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt

- [SDEV3-1718](https://sprucelabsai.atlassian.net/browse/SDEV3-1718)

No, but it does block [SDEV3-1663](https://sprucelabsai.atlassian.net/browse/SDEV3-1663).

### DIFF
--- a/packages/react-heartwood-components/src/components/Forms/components/Autosuggest/Autosuggest.js
+++ b/packages/react-heartwood-components/src/components/Forms/components/Autosuggest/Autosuggest.js
@@ -1,5 +1,7 @@
 // @flow
 import React, { Component } from 'react'
+import { createPortal } from 'react-dom'
+import debounce from 'lodash/debounce'
 import { default as ReactAutosuggest } from 'react-autosuggest'
 import cx from 'classnames'
 import Button from '../../../Button/Button'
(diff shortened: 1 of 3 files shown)


packages/react-heartwood-components/src/components/Forms/components/Autosuggest/Autosuggest.js
```diff
@@ -1,11 +1,15 @@
...
 import React, { Component } from 'react'
+import { createPortal } from 'react-dom'
+import debounce from 'lodash/debounce'
 import { default as ReactAutosuggest } from 'react-autosuggest'
 import cx from 'classnames'
 import Button from '../../../Button/Button'
 import { InputPre, InputHelper } from '../../FormPartials'
 import ClearIcon from '../../../../../static/assets/icons/ic_cancel.svg'
 
+// TODO: Figure out max height setting. Should be based on results length or manual? Both?
```
brandonminch: @jtini will default to you on this but will probably want to implement before merging so we don't add additional tech debt to the implementation in skills/web. My thought is that it could have a default max height (as it does now) but the option to set the height based on the number of preferred visible rows?

PR #521, opened 2019-06-19
fix(PageHeader): Allow setting of additional classnames

gcpantazis:
- In core I'll need this to prevent jank when switching between skill 
view iframes with/without headers
- Cleanup, alphabetize long list of props

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt


gcpantazis:
Actually hold on, what I wanted to do here should be a feature, shouldn't need to add additional classes.

PR #527, opened 2019-07-03
[SDEV3-1889] Fix double JWT authentication issues

kengoldfarb:
Fixes issue where loading a v2 skill in legacy web and then navigating to that skill in heartwood web would cause errors until cookies were cleared.

This _could_ also fix other authentication issues we've seen. Previously, if the jwt was set in the cookie it would use that over the query string which is wrong because the one in cookies might have expired.

### Details

* Prioritize JWTs in the query string over what's in cookies

  * If jwt is found in query string, set it in cookies

* For authentication purposes prefer `jwtV2` over `jwt`

## Type

- [ ] Feature
- [X] Bug
- [ ] Tech debt

- [SDEV3-1889](https://sprucelabsai.atlassian.net/browse/SDEV3-1889)

intersection

### DIFF
--- a/packages/spruce-next-helpers/src/skillskit/next/PageWrapper.js
+++ b/packages/spruce-next-helpers/src/skillskit/next/PageWrapper.js
@@ -120,27 +120,36 @@ const PageWrapper = Wrapped => {
 				renderLocation: renderLocation || 'page'
 			}
 
-			const jwt = query.jwt || getCookie('jwt', req, res)
-
-			// authv1
-			if (jwt) {
-				try {
-					await store.dispatch(actions.auth.go(jwt))
-					setCookie('jwt', jwt, req, res)
-				} catch (err) {
-					debug(err)
-					debug('Error fetching user from jwt')
-				}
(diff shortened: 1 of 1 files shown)


packages/spruce-next-helpers/src/skillskit/next/PageWrapper.js
```diff
@@ -120,27 +120,36 @@ const PageWrapper = Wrapped => {
...
 				try {
 					await store.dispatch(
 						actions.authV2.go(query.jwtV2 || getCookie('jwtV2', req, res))
 					)
-					if (query.jwtV2) {
-						setCookie('jwtV2', query.jwtV2, req, res)
-					}
+				} catch (e) {
+					debug(err)
+					debug('Error fetching user from jwt')
```
ryanrd: are the debug messages swapped? this is for jwtV2 and line 155 is for jwt right?

liquidg3: @kengoldfarb ☝️

kengoldfarb: Fixed

PR #530, opened 2019-07-09
Brandon/feat/truncated list

brandonminch:
## What does this PR do?

Truncated list component to be used to display a set number of items in RecordSelectionList and display an action button to view all results (note: view all w

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt

- [SDEV3-1950](https://sprucelabsai.atlassian.net/browse/SDEV3-1950)

Requires updates to RecordSelectionList to be merged and base branch reset to canary: https://github.com/sprucelabsai/workspace.sprucebot-skills-kit/pull/529/files

Screen Shot 2019-07-09 at 8 00 56 AM
Screen Shot 2019-07-09 at 8 01 04 AM

idk

### DIFF
--- a/packages/react-heartwood-components/src/components/TruncatedList/TruncatedList-story.js
+++ b/packages/react-heartwood-components/src/components/TruncatedList/TruncatedList-story.js
@@ -1,0 +1,29 @@
+// @flow
+import React, { Fragment, Component } from 'react'
+import { storiesOf } from '@storybook/react'
+import {
+	withKnobs,
+	withKnobsOptions,
+	text,
+	boolean,
+	object,
+	select
+} from '@storybook/addon-knobs/react'
+import {
+	userList,
+	userList02,
+	userList03
+} from '../../../.storybook/data/people'
+import TruncatedList from './TruncatedList'
+
+const stories = storiesOf('TruncatedList', module)
+
+stories.addDecorator(
+	withKnobsOptions({
+		escapeHTML: false
+	})
+)
+
+stories.addDecorator(withKnobs)
+
+stories.add('Truncated List', () => <TruncatedList />)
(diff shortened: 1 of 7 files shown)


packages/react-heartwood-components/src/components/TruncatedList/TruncatedList.js
```diff
@@ -0,0 +1,105 @@
...
+import React, { Component, Fragment } from 'react'
+import cx from 'classnames'
+
+import Button from '../Button/Button'
+import EmptyState from '../EmptyState/EmptyState'
+import RecordSelectionList from '../RecordSelectionList/RecordSelectionList'
+
+import type { RecordSelectionListItemProps } from '../RecordSelectionList/RecordSelectionList'
+
+type Props = {
```
jtini: Please add prop descriptions. I know this isn't explicitly in the A/C's, but I think we should include these descriptions any time we add a new component to facilitate collaboration.

packages/react-heartwood-components/src/components/TruncatedList/TruncatedList-story.js
```diff
@@ -0,0 +1,230 @@
...
+						actions={object('actions', [
+							{
+								text: 'Add a location',
+								kind: 'simple'
+							}
+						])}
+					/>
+					<CardBody>
+						<TruncatedListExample
+							header={text('header', 'Locations')}
```
jtini: Can we take `header` out of this example? I wouldn't recommend using it when it's redundant with the card header title. Just wanna make sure this doesn't create confusion.

packages/react-heartwood-components/src/components/TruncatedList/TruncatedList.js
brandonminch: Good call. was an oversight as I was architecting out the props and meant to come back to it.

packages/react-heartwood-components/src/components/TruncatedList/TruncatedList-story.js
brandonminch: 👍

corbanb:
@brandonminch just need the branch updated and its a go.

PR #536, opened 2019-07-10
feat(ModalBody): Allow full bleed modal contents [SDEV3-1814]

gcpantazis:
## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt

- [SDEV3-1814](https://sprucelabsai.atlassian.net/browse/SDEV3-1814)

### DIFF
--- a/packages/react-heartwood-components/src/components/Modal/components/ModalBody/ModalBody.js
+++ b/packages/react-heartwood-components/src/components/Modal/components/ModalBody/ModalBody.js
@@ -10,13 +10,22 @@ type Props = {
 	/** Optional class name */
 	className?: string,
 
+	/** Should the content go to the edge of the modal body? */
+	fullBleed?: boolean,
+
 	/** Optional height of body */
 	height?: string
 }
 
 const ModalBody = (props: Props) => (
 	<div
-		className={cx('modal-body', props.className)}
+		className={cx(
+			'modal-body',
+			{
+				'modal-body__full-bleed': props.fullBleed
+			},
+			props.className
+		)}
 		style={{ ...(props.height ? { height: props.height } : {}) }}
 	>
 		{props.children}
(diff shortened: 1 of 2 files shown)


packages/react-heartwood-components/src/components/Modal/components/ModalBody/ModalBody.js
```diff
@@ -10,13 +10,22 @@ type Props = {
...
 	height?: string
 }
 
 const ModalBody = (props: Props) => (
 	<div
-		className={cx('modal-body', props.className)}
+		className={cx(
+			'modal-body',
+			{
+				'modal-body__full-bleed': props.fullBleed
```
brandonminch: @gcpantazis for modifiers we've typically been using dashes (e.g. `modal-body--full-bleed`)

packages/react-heartwood-components/src/components/Modal/components/ModalBody/ModalBody.js
```diff
@@ -10,13 +10,22 @@ type Props = {
 	/** Optional class name */
 	className?: string,
 
+	/** Should the content go to the edge of the modal body? */
+	fullBleed?: boolean,
```
jtini: Open to y'all's thoughts on this, but do we want to standardize around a format for this type of prop? Like `isWhateverMod`, so in this case `isFullBleed`? I know we're doing it a lot with `isSmall` and `isFullWidth` in other components.

brandonminch: @jtini and @gcpantazis  yeah I think we should keep it consistent with the `isWhatever` format for boolean props. Gets kind of verbose

gcpantazis: Just carrying over convo from slack: 

> Yeah was just thinking about that. Not sure. I like the be-participle prefixed attributes when its a state. Not sure for something like this

> Like `isSmall` not sure I like but `isLoading` I definitely like

> But I think that's because the latter is something stateful... its maybe a good indicator of what is and isn't stateful. :thinking_face:

PR #537, opened 2019-07-10
[SDEV3-2029 ] Split Button component

jtini:
## What does this PR do?

Adds a handy Split Button component that delivers a default action with some sensible related actions. You can see it in action in Storybook with all its various sizes and kinds.

Marking Do Not Merge because this isn't relevant to the current release.

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt

- [SDEV3-2029](https://sprucelabsai.atlassian.net/browse/SDEV3-2029)

## Other components affected
- ButtonGroup was updated to support the ability to highlight an item via class so that items can be highlighted (and triggered) using the keyboard alone. Also, to support full width sizing so that positioning can be easier for the hidden menu.
- **Icon, Button, and ButtonGroup were all converted to TSX**

Nope

Screen Shot 2019-07-10 at 4 57 50 PM


packages/react-heartwood-components/src/components/SplitButton/SplitButton.tsx
```diff
@@ -0,0 +1,253 @@
...
+				/>
+				<Button
+					kind={kind}
+					isSmall={isSmall}
+					className="split-button__actions"
+					icon={{ name: 'keyboard_arrow_down' }}
+					onClick={this.toggleActionsVisibility}
+				/>
+				{typeof document !== 'undefined' &&
+					document &&
```
gcpantazis: I don't **think** you need to check `document` since you verified it wasn't undefined on the previous line

jtini: Got it. Took those out.

PR #538, opened 2019-07-11
fix(PageHeader): Fix element alignment of primaryAction CTA

gcpantazis:
## Type

- [ ] Feature
- [x] Bug
- [ ] Tech debt

Before:
Screen Shot 2019-07-11 at 2 07 10 PM

After:
Screen Shot 2019-07-11 at 2 07 21 PM

### DIFF
--- a/packages/heartwood-components/components/03-structure/page.scss
+++ b/packages/heartwood-components/components/03-structure/page.scss
@@ -12,7 +12,7 @@
 .page__header {
 	padding-top: spacing('base');
 	overflow: hidden;
-	max-height: 200px;
+	max-height: 100vh;
 	opacity: 1;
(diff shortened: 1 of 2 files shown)


packages/heartwood-components/components/03-structure/page.scss
```diff
@@ -110,6 +110,7 @@
 }
 
 .page__header-title-wrapper {
+	align-items: center;
```
jtini: This needs to be wrapped in a media query so that it stacks on mobile and goes inline on larger screens. I think it's safe to use our `small` breakpoint for this.

Screen Shot 2019-07-11 at 1 17 23 PM

gcpantazis: Updated 👍

PR #565, opened 2019-07-24
Update login button href to go to the login page

corbanb:
## Type

- [ ] Feature
- [ ] Bug
- [X] Tech debt

### DIFF
--- a/packages/react-heartwood-components/src/components/Core/components/HeaderPrimary/HeaderPrimary.js
+++ b/packages/react-heartwood-components/src/components/Core/components/HeaderPrimary/HeaderPrimary.js
@@ -238,8 +238,12 @@ export default class HeaderPrimary extends Component<Props, State> {
 						</Fragment>
 					) : (
 						<Fragment>
-							<Button kind="primary" isSmall text="Log In" />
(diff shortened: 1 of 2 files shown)


packages/react-heartwood-components/src/components/Core/components/HeaderPrimary/HeaderPrimary.js
```diff
@@ -238,8 +238,12 @@ export default class HeaderPrimary extends Component<Props, State> {
 						</Fragment>
 					) : (
 						<Fragment>
-							<Button kind="primary" isSmall text="Log In" />
-							<Button isSmall text="Sign Up" />
+							<Button
+								kind="primary"
+								isSmall
+								text="Log In / Sign Up"
+								href="/login"
```
gcpantazis: This should accept a prop, like `loginHref`. I think we'd want to send them to `/`, on core not `/login`.

Also we could move this component over to core, I believe we talked about that before... But for now just the prop is fine.

corbanb: moved to props until moving this out of here into core.

PR #612, opened 2019-08-14
fix(eslint): Remove `browser` env from eslint config.

gcpantazis:
## What does this PR do?

- This caused TS to allow window globals that I don't want, like 
`location`, which given our use of location as a model creates tons of 
bug opportunities. Instead I'll define the exact globals we want to 
allow in that space. 
- Also make sure eslint rules are applied to TS files
- Tested against ai-web to ensure minimal impact

## Type

- [ ] Feature
- [ ] Bug
- [x] Tech debt


gcpantazis:
See that lint is still failing; I'll check back in on this next week

PR #613, opened 2019-08-15
[SDEV3-2306] fix: hidden radio buttons

brandonminch:
## What does this PR do?

Fixes issue where radio buttons are hidden by default because of styles applied to checkbox `indeterminate` state.

## Type

- [ ] Feature
- [X] Bug
- [ ] Tech debt

- [SDEV3-2306](https://sprucelabsai.atlassian.net/browse/SDEV3-2306)

### DIFF
--- a/packages/heartwood-components/components/02-forms/checkbox.scss
+++ b/packages/heartwood-components/components/02-forms/checkbox.scss
@@ -61,7 +61,7 @@
 	}
 
 	&:indeterminate:not(.radio) {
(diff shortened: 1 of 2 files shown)


packages/heartwood-components/components/02-forms/checkbox.scss
```diff
@@ -60,7 +60,7 @@
 		}
 	}
 
-	&:indeterminate {
+	&:indeterminate:not(.radio) {
```
gcpantazis: `:not(input[type="radio"])` so you don't need the class

PR #624, opened 2019-08-21
[SDEV3-2380] feat: Add storybook to spruce-skill

brandonminch:
## What does this PR do?

Adds storybook to `spruce-skill`

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt

- [SDEV3-2380](https://sprucelabsai.atlassian.net/browse/SDEV3-2380)

coolstory


packages/spruce-skill/.storybook/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
```
gcpantazis: In ai-web I'm updating this to:

```
{
	"extends": "../tsconfig.json",
	"compilerOptions": {
		"esModuleInterop": true
	}
}
```

Use the root config, plus add the interop option storybook needs

brandonminch: @gcpantazis not sure why, but I have to also include `"jsx": "react"` here otherwise storybook gets angry.

PR #636, opened 2019-08-28
[SDEV3-2361] Add gql typescript generator

kengoldfarb:
## What does this PR do?

* Implements typescript generator for gql types

`yarn gqlToTS`

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt

- [SDEV3-2361](https://sprucelabsai.atlassian.net/browse/SDEV3-2361)


kengoldfarb:
We already had a `gql.ts` file that was getting overwritten by the codegen.

I've reverted that file and changed the generated filename to `gql-generated.ts` to help avoid future confusion

PR #640, opened 2019-08-29
[SDEV3-2361] Upgrade graphql version

kengoldfarb:
## What does this PR do?

`graphql-js` is now shipping w/ typescript definitions so the `@types` definition is no longer needed

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt

## What are the relevant tickets?

- [SDEV3-2361](https://sprucelabsai.atlassian.net/browse/SDEV3-2361)

PR #641, opened 2019-08-30
[SDEV3-2361] Strongly typed "config"

kengoldfarb:
## What does this PR do?

* Upgrade `config` package to latest version

* Adds strong typing to `config`
  * Works w/ `config.<start typing>` - vscode intellisense support

* Moves majority of config vars out to `spruce-skill-server` with descriptions
  * Greatly simplifies `spruce-skill` default config

* Splits out acl, eventContract, etc. into separate config files

* Upgrade graphql version which just started shipping with types (also in standalone separate PR: https://github.com/sprucelabsai/workspace.sprucebot-skills-kit/pull/640)

* Fix issue where TS gql resolvers were not loaded when running locally

* Removes flow types

* Next.js config loads env vars directly from `client.json` which is built with `yarn build:config` (also part of `yarn build`)
  * _Improvement:_ We may need to add file watching to automatically regenerate this file for local development

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt

- [SDEV3-2361](https://sprucelabsai.atlassian.net/browse/SDEV3-2361)


packages/spruce-skill-server/config/default.ts
```diff
@@ -0,0 +1,288 @@
+/* eslint-disable @typescript-eslint/camelcase */
+/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
+// https://github.com/lorenwest/node-config/wiki/Configuration-Files
+import fs from 'fs'
+import baseErrors from './errors'
+
+export default function SpruceConfig<
+	aclType,
```
liquidg3: Since these are generated by doing a `typeof` are they interfaces or types? Should we capitalize on them? We have historically capitalized types (thinking flow), not sure if we are carrying that forward.

packages/spruce-skill-server/config/default.ts
```diff
@@ -0,0 +1,288 @@
...
+			enabled: process.env.DB_ENABLED === 'true',
+			runMigrations: process.env.DB_MIGRATIONS === 'true',
+			modelsDir: `${baseDirectory}/models`,
+			migrationsDir: `${baseDirectory}/migrations`,
+			// Additional sequelize options
+			options: {
+				logging: process.env.ORM_LOGGING === 'true' ? console.log : false
+			}
+		},
+		bodyParserOptions: {
```
liquidg3: can we use `typeof Koa.BodyParser` or something here?

liquidg3: Will https://www.npmjs.com/package/@types/koa-bodyparser work?

packages/spruce-skill-server/config/default.ts
```diff
@@ -0,0 +1,288 @@
...
+			// passthrough to https://github.com/koajs/bodyparser
+			jsonLimit: '5mb'
+		},
+		nextConfig: {
+			dir: `${process.cwd()}/interface`,
```
liquidg3: same here, possible to use a next config interface?

packages/spruce-skill-server/gql/router.ts
```diff
@@ -100,12 +98,16 @@ export default (
...
 					const code = e.message
 					let formattedError: Record<string, any> = {}
-
+					const errors = config.errors
+					// @ts-ignore
 					if (errors[code]) {
 						formattedError = {
 							name: code,
+							// @ts-ignore
 							code: errors[code].code,
```
liquidg3: For situations like this, is it practical to put a `get` method on both errors and config like this:

```
config.get(key: [K as keyof config]): config[k]
```
I can't remember the exact syntax, but i did something like this before where it only let you pass keys and it wokred pretty well.

packages/spruce-skill-server/config/default.ts
kengoldfarb: Yep, makes sense

packages/spruce-skill-server/config/default.ts
kengoldfarb: Yeah, the issue w/ setting an interface on an object literal is that you lose the exact type...so in this case autocomplete would just show me the interface type when I start typing `config.bodyParserOptions.` instead of showing me exactly what's defined.

Same thing w/ errors and settings...you'll see the TODO comment in those files.  In the future once TS catches up and supports exact types we should be able to do that

PR #676, opened 2019-09-26
Added `friendlyDuration` and `friendlyCurrency` utils

rickycpadilla:
## What does this PR do?
Added `friendlyDuration` and `friendlyCurrency` utils

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt


packages/spruce-utils/lib/il8n.tsx
```diff
@@ -0,0 +1,35 @@
+/** Returns a string of hours and minutes. Ex: "4hr 30 min" */
+export const friendlyDuration = (seconds: number, lang?: any): string => {
```
kengoldfarb: Can we use moment for this instead?

https://momentjs.com/docs/#/i18n/

packages/spruce-utils/lib/il8n.tsx
```diff
@@ -0,0 +1,35 @@
...
+		if (duration.length > 0) {
+			duration += ' '
+		}
+		duration += `${remainingMin}${minString || 'min'}`
+	}
+	return duration
+}
+
+/** Converts a numnber of US cents into a formatted currency string. US Dollars by default. Ex: "$23.99" */
+export const friendlyCurrency = (
```
kengoldfarb: Same question here...can we use a library instead of rolling our own?

https://currency.js.org/ looks good though I haven't used it myself

rickycpadilla: Ooo, yeah that looks better. Thanks!

packages/spruce-utils/lib/il8n.tsx
rickycpadilla: Unfortunately this library doesn't have what we need for specifying a duration format in different languages :(

packages/spruce-next-helpers/src/skillskit/helpers/lang.ts
```diff
@@ -0,0 +1,115 @@
...
+		throw Error(`Translation missing key ${key}`)
+	},
+	/** Converts a numnber of cents into a formatted currency string. Ex: "$23.99" */
+	friendlyCurrency(cents: number, location: ICoreLocation): string {
+		/* TODO: Use these props once added to Location
+		const currency: string = location.currecy || 'USD'
+		const locale: string = location.locale || 'en-US'
+		const precision: number = location.currencyMinorUnits || 2
+		*/
+		console.log('Location passed to friendlyCurrency: ', location.id)
```
gcpantazis: Remove log; add an eslint ignore next-line for `location` if it complains about that, should be a fix suggestion in vscode.

packages/spruce-next-helpers/src/skillskit/helpers/lang.ts
```diff
@@ -0,0 +1,115 @@
...
+			.replace('.00', '')
+		return friendlyCurrency
+	},
+	/** Returns a duration string of largest time increments from numer of seconds. Ex: `friendlyDuration(3452626, 2)` returns "1mo 8d" */
+	friendlyDuration(seconds: number, location: ICoreLocation): string {
+		interface IDuration {
+			value: number
+			label: string
+		}
+		console.log('Location passed to friendlyDuration: ', location.id)
```
gcpantazis: Ditto, remove

packages/spruce-next-helpers/src/skillskit/helpers/lang.ts
```diff
@@ -0,0 +1,115 @@
...
+		} catch (err) {
+			console.info('No lang override specified.')
+		}
+	},
+	mixin(lang, overrides) {
+		this.lang = lang
+		this.overrides = overrides
+	},
+	getText(key: string, context: Record<string, any> = {}) {
+		const translations: Record<string, any> = {
```
gcpantazis: This will create an in-memory copy of the entire lang, plus overrides every time you call getText; better would be to look inside context, then overrides, then lang for key.

packages/spruce-next-helpers/src/skillskit/helpers/lang.ts
```diff
@@ -0,0 +1,115 @@
+import * as moment from 'moment'
```
gcpantazis: Could this move to `spruce-utils`? This is perhaps more generic than just `next`

packages/spruce-next-helpers/src/skillskit/helpers/lang.ts
```diff
@@ -0,0 +1,115 @@
...
+import Dinero from 'dinero.js'
+
+import { ICoreLocation } from '@sprucelabs/spruce-types'
+
+export default {
+	lang: {},
+	overrides: {},
+	async configure(langDir: string) {
```
gcpantazis: As far as making this more generic, what about passing in `lang` and `overrides` as objects, and defer to the individual usage to require and provide the data, rather than `require`ing here? Would also make this more testable I think.

kengoldfarb:
Closing. Task to create RFC: https://sprucelabsai.atlassian.net/browse/SDEV3-2529

PR #677, opened 2019-09-26
feat(Typescript): Upgrade to 3.6.3

gcpantazis:
- Upgrades to TS 3.6.3, which necessitated an upgrade to our TS+eslint integration.
- Fixes mostly for `prefer-const`.
- Fix a few typing issues that 3.6.3 was able to flag that 3.4.x missed.

### DIFF
--- a/packages/eslint-config-spruce/index.js
+++ b/packages/eslint-config-spruce/index.js
@@ -37,6 +37,11 @@ module.exports = {
 					{ allow: ['^(can_|skill_can_)'] }
 				],
 				'@typescript-eslint/no-empty-interface': 0,
+				// TODO: Remove this if we can; it isn't a good rule to squash.
+				// Sometimes this is fine, but sometimes it masks a compile error.
+				'@typescript-eslint/ban-ts-ignore': 0,
+				'@typescript-eslint/no-empty-function': 0,
+				'@typescript-eslint/explicit-function-return-type': 0,
 				'@typescript-eslint/interface-name-prefix': [2, 'always'],
 				'@typescript-eslint/no-explicit-any': 0,
 				'@typescript-eslint/member-delimiter-style': [
@@ -50,12 +55,6 @@ module.exports = {
 							delimiter: 'semi',
 							requireLast: false
 						}
-					}
-				],
-				'@typescript-eslint/explicit-function-return-type': [
-					'error',
-					{
-						allowExpressions: true
 					}
 				],
 				'@typescript-eslint/member-ordering': [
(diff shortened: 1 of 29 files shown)


packages/eslint-config-spruce/index.js
```diff
@@ -52,12 +57,6 @@ module.exports = {
 						}
 					}
 				],
-				'@typescript-eslint/explicit-function-return-type': [
```
gcpantazis: @kengoldfarb It was easier to just disable this... I figure individual projects can opt into `strict` TS, which would catch all of these issues.

packages/react-heartwood-components/src/components/Card/components/CardBuilder.tsx
```diff
@@ -1,34 +1,28 @@
...
+import Toast, { IToastProps } from '../../Toast/Toast'
+import Card from '../Card'
+import CardBody from './CardBody'
+import CardFooter from './CardFooter'
+import CardHeader, { ICardHeaderProps } from './CardHeader'
+import OnboardingCard, { IOnboardingCardProps } from './OnboardingCard'
+import Scores from './Scores'
 
-export interface ICardBuilderFooter extends Omit<IHWCardFooter, 'buttonGroup'> {
+export interface ICardBuilderFooter extends IHWCardBuilderFooter {
```
gcpantazis: @liquidg3 This is the type change I mentioned in Slack

packages/spruce-node/index.ts
```diff
@@ -652,7 +652,7 @@ export default class Sprucebot {
 
 		// not found, create it
 		if (!meta) {
-			meta = await this.createMeta(key, value, Array.from(arguments)[2])
```
gcpantazis: @liquidg3 @kengoldfarb can you sanity check this? At a blush `Array.from(arguments)[2]` would always get options...  but maybe I'm missing something.

packages/spruce-skill-server/lib/GraphQLSubscriptionServer.ts
```diff
@@ -102,6 +102,9 @@ export default class GQLSubscriptionServer {
 		const token = authorizationHeader.replace('JWT ', '')
 		const decoded = jwt.verify(
 			token,
+			// TODO: config's type doesn't describe this API correctly...
```
gcpantazis: @kengoldfarb Not sure what the deal is here, but noticed that the types file hasn't been updated in npm in a few years. Possibly just totally wrong. We forked this right?

packages/react-heartwood-components/src/components/Card/components/CardBuilder.tsx
liquidg3: fixed.

packages/spruce-node/index.ts
liquidg3: yes, this was legacy when there was magic on the signature of that method. your fix is 100%.

packages/spruce-skill-server/lib/GraphQLSubscriptionServer.ts
kengoldfarb: I think we should now be able to just reference it directly: `config.API_KEY` and have it strongly typed

PR #679, opened 2019-09-27
Prerelease/calendar events interfaces

liquidg3:
## What does this PR do?

Makes some interface changes for event listeners.

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt

NA

NA

NA

NA

### DIFF
--- a/packages/spruce-skill-server/factories/sequelize.ts
+++ b/packages/spruce-skill-server/factories/sequelize.ts
@@ -175,5 +175,5 @@ export default (
 		}
 	}
 
-	ctx[key] = { models, sequelize, sync, Sequelize }
+	ctx[key] = { models, sequelize, sync }
 }
--- a/packages/spruce-skill-server/factories/sequelize.ts
+++ b/packages/spruce-skill-server/factories/sequelize.ts
@@ -175,5 +175,5 @@ export default (
 		}
 	}
 
-	ctx[key] = { models, sequelize, sync }
+	ctx[key] = { models, sequelize, sync, Sequelize }
 }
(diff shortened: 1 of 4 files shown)


packages/spruce-skill-server/factories/sequelize.ts
```diff
@@ -175,5 +175,5 @@ export default (
 		}
 	}
 
-	ctx[key] = { models, sequelize, sync }
+	ctx[key] = { models, sequelize, sync, Sequelize }
```
kengoldfarb: It seems like we shouldn't need to set `Sequelize` here since we can just import it elsewhere if we need it

packages/spruce-skill-server/interfaces/ctx.ts
```diff
@@ -22,6 +22,7 @@ export interface ISpruceContext<
 	db: {
 		models: ISkillModels
 		sequelize: Sequelize
+		Sequelize: typeof Sequelize
```
kengoldfarb: Necessary?  see above

packages/spruce-skill-server/factories/sequelize.ts
liquidg3: Inside of booking, we invoke `ctx.db.sequelize.literal(...` and get a "Calling method on static function".

Booking will have a corresponding PR with, `ctx.db.Sequelize.literal(...`.

kengoldfarb: Yeah, since we're not initializing anything and it's just the static methods I was thinking we'd just import it:

```
import { literal } from 'sequelize'

ctx.db.models.User.findOne({
  where: literal(...)
})
```

PR #690, opened 2019-10-08
feat: loading on lists

liquidg3:
## What does this PR do?

brings in a classy loading state for list items

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt

NA

NA

Oct-07-2019 19-27-57


packages/spruce-types/src/gql/List.gql
```diff
@@ -94,6 +94,9 @@ type ListItem {
 	"Highlight title, subtitle, note with warning colors"
 	warnings: ListItemWarningConfig
 
+	"Is this list item in a loading state?"
+	isLoading: Boolean
```
gcpantazis: This feels like something that should just be on the component (i.e. just on `IListProps`). I'm not sure if the API would need to know about it. But maybe you have a use-case in mind here?

If so I'm just curious how you'd get _out_ of the loading state if it was set by the API 🤔

liquidg3: good point!

PR #692, opened 2019-10-09
fix(GQL): Convert gql util client to TS [SDEV3-509]

gcpantazis:
Just a prepwork PR for adding GQL subscription helpers here

## Type

- [ ] Feature
- [ ] Bug
- [x] Tech debt

- [SDEV3-509](https://sprucelabsai.atlassian.net/browse/SDEV3-509)

### DIFF
--- a/packages/spruce-utils/lib/graphql.ts
+++ b/packages/spruce-utils/lib/graphql.ts
@@ -2,7 +2,10 @@ import { compact, get } from 'lodash'
 import { ApolloClient } from 'apollo-client'
 import { ApolloLink } from 'apollo-link'
 import { createHttpLink } from 'apollo-link-http'
-import { InMemoryCache } from 'apollo-cache-inmemory'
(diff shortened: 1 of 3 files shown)


packages/spruce-utils/lib/graphql.ts
```diff
@@ -91,30 +97,40 @@ export class GraphQLClient {
 				}
 			})
 
+			// @ts-ignore
```
gcpantazis: @kengoldfarb `wsLink.subscriptionClient` is a private field, so TS doesn't like this. Just ignoring that one, but if there's a better public method that'd be a good follow-up PR

kengoldfarb: Hmm, that makes me think maybe we're doing it wrong...I'm not seeing any open issues related to that TS error.

But agreed it makes sense as a separate PR

PR #706, opened 2019-10-20
HW components can take both generated props and standard props

liquidg3:
## What does this PR do?

Lets you use heartwood components manually and through gql endpoints without warnings/errors.

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt

- [SDEV3-#](https://sprucelabsai.atlassian.net/browse/SDEV3-#)

### DIFF
--- a/packages/react-heartwood-components/src/components/Card/components/CardHeader.tsx
+++ b/packages/react-heartwood-components/src/components/Card/components/CardHeader.tsx
@@ -20,8 +20,15 @@ export interface ICardHeaderProps
 	contextMenu?: IContextMenuProps | null
 }
 
-const CardHeader = (props: ICardHeaderProps): React.ReactElement => {
-	const { title, labelText, labelIcon, actions, contextMenu } = props
+const CardHeader = (
+	props: ICardHeaderProps | IHWCardHeader
+): React.ReactElement => {
+	const propsComp = props as ICardHeaderProps
+	const propsGql = props as IHWCardHeader
+
+	const { title, labelText, actions, contextMenu } = propsGql
+	const { labelIcon } = propsComp
+
 	return (
 		<div className="card__header">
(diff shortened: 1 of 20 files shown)


packages/react-heartwood-components/src/components/Card/components/CardHeader.tsx
```diff
@@ -20,8 +20,15 @@ export interface ICardHeaderProps
 	contextMenu?: IContextMenuProps | null
 }
 
-const CardHeader = (props: ICardHeaderProps): React.ReactElement => {
-	const { title, labelText, labelIcon, actions, contextMenu } = props
+const CardHeader = (
+	props: ICardHeaderProps | IHWCardHeader
+): React.ReactElement => {
+	const propsComp = props as ICardHeaderProps
```
gcpantazis: Per our convo, let do `reactHeartwoodProps` instead of `propsComp` for `ICardHeaderProps` (the props you've extended on top of IHWCardHeader). propsGql should be `commonProps`.

Goal here is just clarity of what goes where and why without having too much context of the underlying structure

packages/react-heartwood-components/src/components/Card/components/CardHeader.tsx
```diff
@@ -20,8 +20,15 @@ export interface ICardHeaderProps
...
+	const propsGql = props as IHWCardHeader
+
+	const { title, labelText, actions, contextMenu } = propsGql
+	const { labelIcon } = propsComp
```
gcpantazis: Per convo, everything you've changed as a part of `ICardHeaderProps` should go in here; if there's a substantial divergence in the two types we can do `foo as commonFoo` and `foo as reactHeartwoodFoo` etc. to craft the logic, which would again preserve DX comprehension w/o a ton of mental leaps :D

PR #713, opened 2019-10-22
new upcoming cal event kind, fixed icon story

liquidg3:
## What does this PR do?

Makes changes so calendar events can be styled better.

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt

NA


packages/spruce-types/src/generated/api-gql.ts
```diff
@@ -1553,6 +1557,11 @@ export type ICoreGQLGetSettingsResponse = {
 	settings?: Maybe<Scalars['JSON']>
 }
 
+export type ICoreGQLGetSkillTokenResponse = {
```
liquidg3: looks like these didn't get generated from a previous release.

PR #726, opened 2019-10-29
Add optional prop for legacy stylesheet

adamgoth:
## What does this PR do?

Looks for a prop in initial props on components wrapped in `PageWrapper` called `legacyStylesheet`. With this change, the stylesheet specified in the `.env` file as `SKILL_STYLESHEET` will be ignored unless the `legacyStylesheet` is explicitly passed as `true`.

Before merging, we need to consider if we should update the config name of `SKILL_STYLESHEET` to indicate its use now as a legacy stylesheet. We should also consider all the pages we will need to update and pass this prop in as true.


packages/spruce-next-helpers/src/skillskit/next/_document.js
```diff
@@ -56,20 +56,22 @@ export default class MyDocument extends Document {
...
 					<meta name="viewport" content="width=device-width, initial-scale=1" />
-					{publicRuntimeConfig && publicRuntimeConfig.SKILL_STYLESHEET && (
-						<link
-							href={publicRuntimeConfig.SKILL_STYLESHEET}
-							rel="stylesheet"
-							type="text/css"
-							charSet="UTF-8"
-						/>
-					)}
+					{legacyStylesheet &&
```
gcpantazis: This shouldn't be in `_document.js`, since _theoretically_ you could client-side nav between a legacy and a heartwood page. It should move over to `_app.js`, which is rendered per-page client-side

packages/spruce-next-helpers/src/skillskit/next/PageWrapper.js
```diff
@@ -114,7 +115,8 @@ const PageWrapper = Wrapped => {
 			store,
 			res,
 			req,
-			renderLocation
+			renderLocation,
+			legacyStylesheet
```
gcpantazis: Do you need to do something with this? 🤔

adamgoth:
> Could you also implement an example page showing this as working?

@gcpantazis I spoke with @liquidg3 this morning about adding an example to a page and he said not to add one. Do you either of you feel strongly about it? It's a trivial add on my end but not sure what consensus is.

gcpantazis:
Eh it's a good point now that I think of it... @liquidg3 is probably thinking that we don't want to expose this as a normal use-case to a skills dev. So long as you verified it I think we're fine.

packages/spruce-skill/interface/pages/_app.js
```diff
@@ -54,6 +58,21 @@ export default class MyApp extends App {
 
 		return (
 			<Container>
+				{pageProps.initialProps.legacyStylesheet && (
+					<Head>
+						<meta
```
gcpantazis: Delete this since you also have it in document.js

adamgoth: In the next docs it shows having the meta tag twice with the same key to avoid duplications https://nextjs.org/docs#populating-head

PR #737, opened 2019-11-01
fix(npm-Cookies): [v7] SameSite: None + cookie upgrade [SDEV3-2622]

gcpantazis:


gcpantazis:
Don't need to update v7 anymore, closing

PR #756, opened 2019-11-05
fix(npm-Cookies): v8.x: Update to version supporting sameSite [SDEV3-2622]

gcpantazis:
- [SDEV3-2622](https://sprucelabsai.atlassian.net/browse/SDEV3-2622)

### DIFF
--- a/packages/spruce-next-helpers/src/skillskit/next/PageWrapper.js
+++ b/packages/spruce-next-helpers/src/skillskit/next/PageWrapper.js
@@ -20,9 +20,9 @@ const setCookie = (named, value, req, res) => {
 			secure: true,
 			httpOnly: false
 		})
-		return cookies.set(named, value)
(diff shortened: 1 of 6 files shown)


packages/spruce-next-helpers/src/skillskit/next/PageWrapper.js
```diff
@@ -20,9 +20,9 @@ const setCookie = (named, value, req, res) => {
 			secure: true,
 			httpOnly: false
 		})
-		return cookies.set(named, value)
+		return cookies.set(named, value, { sameSite: 'None' })
```
kengoldfarb: Does the case of `'None'` matter? It's uppercase here but lowercase in `auth.js`

gcpantazis: Nope it doesn't, but I'll make it upper-case for consistency.

PR #762, opened 2019-11-06
Add a legacy stylesheet var to default

adamgoth:
## What does this PR do?

SKILL_STYLESHEET defaults to heartwood-components.css if it is not specified. This is what devs will want in almost all cases. However we will leave the var in case they want to override it with a different stylesheet.

A new var LEGACY_SKILL_STYLESHEET has been added and will be the stylesheet that gets pulled in when a user passes an initial prop of `legacyStylesheet: true` on a page.

This has been tested in booking skill and seems to fix the CSS override bug that currently exists.


kengoldfarb:
These would also need to be updated to support this var right?
1. https://github.com/sprucelabsai/workspace.sprucebot-skills-kit/blob/a4417107d36593e9c38ce1d01ff93721d8c9f997/packages/spruce-skill/interface/pages/_app.js#L61

2. https://github.com/sprucelabsai/workspace.sprucebot-skills-kit/blob/a4417107d36593e9c38ce1d01ff93721d8c9f997/packages/spruce-skill/config/default.ts#L43

adamgoth:
> These would also need to be updated to support this var right?
> 
> 1. https://github.com/sprucelabsai/workspace.sprucebot-skills-kit/blob/a4417107d36593e9c38ce1d01ff93721d8c9f997/packages/spruce-skill/interface/pages/_app.js#L61
> 2. https://github.com/sprucelabsai/workspace.sprucebot-skills-kit/blob/a4417107d36593e9c38ce1d01ff93721d8c9f997/packages/spruce-skill/config/default.ts#L43

Good catch. Updated both places.

PR #765, opened 2019-11-08
big search move

liquidg3:
## What does this PR do?

bring big search interfaces in with other event interffaces

## Type

- [ ] Feature
- [ ] Bug
- [ ] Tech debt

- [SDEV3-#](https://sprucelabsai.atlassian.net/browse/SDEV3-#)


kengoldfarb:
TODO: We'll need to move these definitions to the API event contract which will get pulled in by skills as part of our strong event types initiative

PR #772, opened 2019-11-19
[SDEV3-2751] Shared TS types for all events across skills and core

kengoldfarb:
## What does this PR do?

Implementation based on RFC: https://github.com/sprucelabsai/spruce-rfcs/pull/6

Pulls TS event names, payload, and body types from API

See [the generated typescript file](https://github.com/sprucelabsai/workspace.sprucebot-skills-kit/compare/prerelease/SDEV3-2751-shared-event-signatures-squash?expand=1#diff-9d9093e6ac83e3f22c84820700e560cb) that contains namespaced events w/ interfaces

On boot, after syncing skill settings with the api, it will pull down the latest types into the `server/interfaces/events-generated.ts` file.

In CI we could run a comparison and fail builds if we get a difference from the generated file to ensure that event contracts stay in sync between skills and core

## Type

- [X] Feature
- [ ] Bug
- [ ] Tech debt

- [SDEV3-2751](https://sprucelabsai.atlassian.net/browse/SDEV3-2751)


kengoldfarb:
@liquidg3 yep, i agree about moving it all under a single event namespace like that.  I also want to limit the event types to skills in the event contract. I worry if we pull back everything we'll get a bunch of conflicts when running locally since we'll all have different skill installed.

As for calling it "Body", I think it makes sense in the context of handling an event since we're literally writing `ctx.body={...}`. Maybe we go more verbose like: `SpruceEvents.booking.CreateAppointment.IResponseBody`?

kengoldfarb:
@liquidg3 I've moved event parts under a common namespace, set it up to sync only types the skill needs or requests, and created an example event definition and handler.

kengoldfarb:
1. It's defined like this so that we can have nested arrays / objects. For example:

```
childNames: {
	description: "A random array of this person's children's names",
	type: SpruceDataTypes.Array,
	items: {
		firstName: {
			type: SpruceDataTypes.String,
			description: 'the chosen first name'
		},
		middleName: {
			type: SpruceDataTypes.String,
			description: 'the chosen middle name'
		},
		possibleNames: {
			type: SpruceDataTypes.Array,
			description: 'array of names that were not chosen',
			items: {
				firstName: {
					type: SpruceDataTypes.String,
					description: 'a first name that was ruled out'
				},
				middleName: {
					type: SpruceDataTypes.String,
					description: 'a middle name that was ruled out'
				}
			}
		}
	}
}
```

2. Yeah, we can change it. I was just being extra explicit

kengoldfarb:
@liquidg3 as discussed I've now dropped the `SpruceDataTypes.Array` in favor of `isArray: true` and changed `items` to `definition

PR #776, opened 2019-11-26
Keep sidebar at 360px until screen is 360px or less

adamgoth:


packages/heartwood-components/components/99-web/sidebar.scss
```diff
@@ -6,8 +6,11 @@
 	background-color: $c-grey-00;
 
 	&.sidebar--large {
-		width: 100%;
-		max-width: rem(360);
+		width: rem(360);
+		@include lte(360px) {
```
adamgoth: @jtini @gcpantazis our 'small' breakpoint is 500px, I'm wondering if we want to add a 'xsmall' one at 360px or just have this one off

jtini: I'd do this as a one-off, since the breakpoint is contextual to the component.

I'll add as my two cents, and just for consideration, I'm not 100% on board with embracing max-width media queries. We run the risk of optimizing for desktop and treating mobile as an afterthought. Up to y'all of course, just mentioning this based on my own experience. I've always found mobile-first and min-width queries seem to be a more efficient workflow.

adamgoth: I agree that the baseline should be smaller, mobile-friendly sizes first, and adjust as screens get larger. George and I were talking yesterday about how our CSS is in need of a refactor/clean up. We should really only be using media queries or mixins like gt and lte, not a mix of both.

PR #800, opened 2020-01-29
Add org id to user query on auth

adamgoth:
## What does this PR do?

Add org id to user query on auth

This fixes issue with null returning for certain properties on user

## Type

- [ ] Feature
- [x] Bug
- [ ] Tech debt


packages/spruce-skill/config/auth.ts
```diff
@@ -10,10 +10,11 @@ const gql = (options: {
 }): string => `
 {
 	${
-		options.userId
+		options.userId && options.organizationId
```
kengoldfarb: With further thought I think we need to adjust this to include the organizationId if it's set but not require it (and still load the `User` like before).  I'm thinking about a skill that enhances a page outside an organization like my personal profile page.

adamgoth: Let me know what you think about the change I just pushed. Basically added a separate query for when there is a user id but no org id and removed UserLocations and UserOrganizations from that query since they come back null without an org id.

ryanrd:
@adamgoth / @kengoldfarb  should this be merged (after fixing the out of date issue)?

PR #809, opened 2020-02-12
Remove windowOrDocument method

adamgoth:


packages/react-sprucebot/lib/skillskit/index.js
```diff
@@ -391,7 +375,7 @@ var skill = {
 };
 
 if (typeof window !== 'undefined') {
-  skill.windowOrDocument().addEventListener('message', skill.handleIframeMessage.bind(skill));
+  skill.window.addEventListener('message', skill.handleIframeMessage.bind(skill));
```
liquidg3: drop `skill`

PR #812, opened 2020-02-17
[SDEV3-3054] fix user menu toggle

adamgoth:
## What does this PR do?

Fixes issue where certain tappable area near avatar would result in user menu immediately closing

## Type

- [ ] Feature
- [x] Bug
- [ ] Tech debt

- [SDEV3-3054](https://sprucelabsai.atlassian.net/browse/SDEV3-3054)

### DIFF
--- a/packages/react-heartwood-components/src/components/Core/components/HeaderPrimary/HeaderPrimary.js
+++ b/packages/react-heartwood-components/src/components/Core/components/HeaderPrimary/HeaderPrimary.js
@@ -81,13 +81,10 @@ export default class HeaderPrimary extends Component<Props, State> {
 	userMenuRef: any
 
 	hideUserMenu = (e: Event) => {
-		console.log(e)
-		console.log(this.userMenuRef)
-		console.log(this.userMenuRef.current)
 		if (
 			e.key === 'Escape' ||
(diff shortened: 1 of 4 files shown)


packages/react-heartwood-components/src/components/Core/components/HeaderPrimary/HeaderPrimary.js
```diff
@@ -83,7 +83,7 @@ export default class HeaderPrimary extends Component<Props, State> {
 		if (
 			e.key === 'Escape' ||
 			e.target.contains(this.ref) ||
-			(e.type === 'click' && e.target.id !== 'avatar')
+			(e.type === 'click' && !e.target.closest('#userMenuToggle'))
```
kengoldfarb: Couple questions...
1. Is there ever a scenario where we'd have multiple `UserMenu` components on a page?
2. Is there a more React-y way of doing this w/ refs instead of looking at the DOM element directly?

adamgoth: It's hard to imagine a scenario where there are multiple UserMenu components on the page, but not knowing the answer to that question, I agree using a ref is more suitable than an id.

PR #816, opened 2020-02-21
[SDEV3-3054] User menu iframe click

adamgoth:
## What does this PR do?

Detect click into iframe so that user menu closes if open


adamgoth:
@liquidg3 let me know what you're thinking for the check in place of  `document.activeElement === document.getElementsByClassName('iframe-page-iframe')[0]`

PR #819, opened 2020-03-05
CardBuilder: Clean up types for better adherence to shared types

gcpantazis:
## What does this PR do?
- Formerly we tried omitting and overwriting props from spruce-types due to slight differences in the React vs. data-driven APIs.
- Since then we've upped our understanding of this type of extension, and it's less necessary. Most specifically I'm doing the correct union-type to allow extension of `children` onto CardBody, which was the big hold-up.
- It's important that builders do not modify our core types going forward, as we're using those as the source of truth. Variance here will lead to bugs, etc.

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt

### DIFF
--- a/packages/spruce-types/src/gql/types/Button.gql
+++ b/packages/spruce-types/src/gql/types/Button.gql
@@ -45,7 +45,7 @@ type Button implements ActionExecutor {
 	type: ButtonTypes
 
 	"Otherwise unspecified attributes that will be applied to the underlying button element"
-	HTMLAttributes: JSON
+	htmlAttributes: JSON
 
 	"Set true to disable the button"
 	isDisabled: Boolean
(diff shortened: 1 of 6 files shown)


kengoldfarb:
@gcpantazis I switched the base to `v9-canary` which is what we should use now as we prepare for v10

packages/spruce-types/src/gql/types/Button.gql
```diff
@@ -44,6 +44,9 @@ type Button implements ActionExecutor {
 	"Type attribute for HTML button element. Defaults to 'button'."
 	type: ButtonTypes
 
+	"Otherwise unspecified attributes that will be applied to the underlying button element"
+	HTMLAttributes: JSON
```
liquidg3: hey @gcpantazis, why not `htmlAttributes`? I can't recall any places that we use abbreviations, but I do have a memory of seeing `json` in some place and it was lower cased.

liquidg3: Actually, lets afterparty on this and decide.

PR #821, opened 2020-03-06
[SDEV3-3090] sprucebot avatar

liquidg3:
## What does this PR do?

Introduces the new Sprucebot avatar, with one state of mind (chilling). Future states will be supplied by designer soon.

## Type

- [x] Feature
- [ ] Bug
- [ ] Tech debt

- [SDEV3-3090](https://sprucelabsai.atlassian.net/browse/SDEV3-3090)


packages/spruce-types/src/gql/types/SprucebotAvatar.gql
```diff
@@ -0,0 +1,32 @@
...
+	"When submitting a form or running some other operation"
+	thinking
+	"Asking a question or leaning new things"
+	inquisiting
+	"Something has finished and we're so excited!"
+	jamming
+}
+
+enum SprucebotAvatarSize {
+	"Inline with text"
```
gcpantazis: Can these be more prescriptive/documentation-focused, i.e. "This size should be used inline with text" "This size is best used for subheadings" etc

packages/spruce-types/src/gql/types/SprucebotAvatar.gql
```diff
@@ -0,0 +1,32 @@
+"Sprucebot's current state of mind"
+enum SprucebotAvatarStateOfMind {
+	"For when nothing in particular is happening"
```
gcpantazis: Let's make sure the voice of the docs here is consistent for all options (i.e. all should start with "For use when..." to be more prescriptive or "Nothing is happening right now" if more descriptive)

packages/spruce-types/src/gql/types/SprucebotAvatar.gql
```diff
@@ -0,0 +1,32 @@
...
+	small
+	"A subheading"
+	medium
+	"A heading"
+	large
+}
+
+"Your friendly neighborhood Sprucebot Avatar!"
+type SprucebotAvatar {
+	"ID for view caching"
```
gcpantazis: Eh, don't think we need this. IDs are used for other reasons and people will generally understand what an ID is anyway.

gcpantazis: Alternatively it can just be generic "A unique identifier for this record"

packages/spruce-types/src/gql/types/SprucebotAvatar.gql
```diff
@@ -0,0 +1,32 @@
+"Sprucebot's current state of mind"
+enum SprucebotAvatarStateOfMind {
```
gcpantazis: If this is answering the question "How does Sprucebot feel?" shouldn't the options be emotions? i.e. `calm`, `contemplative`, `curious`, `productive` respectively for the four options here.

gcpantazis: This feels like a nit, but it also seems kind of critical in getting the grokability of this on point

packages/spruce-types/src/gql/types/SprucebotAvatar.gql
liquidg3: @gcpantazis - I had been making ID's required for all components so we never run the risk of showing a "you must set a key on a component when looping" or whatever that error is.

gcpantazis: Sorry, I meant keep `id` but drop the specific comment "ID for view caching" since IDs are more generic than that.

